PR #6, opened 2017-11-14
Add support for aiobotocore

### LINKED ISSUE #5: Supporting aiobotocore
Back again with more async stuff :smile:

I've been using the aiobotocore library which essentially wraps botocore and uses aiohttp to perform async http requests.

I was looking into patching parts of it so the SDK creates subsegments when queries to AWS are made.
I've gotten it to work but its a bit messy. 

The following is based off the botocore patcher. (The injecting headers bit is exactly the same bar a different class needs patching)

```python
    wrapt.wrap_function_wrapper(
        'aiobotocore.client',
        'AioBaseClient._make_api_call',
        _xray_traced_aiobotocore,
    )

async def _xray_traced_aiobotocore(wrapped, instance, args, kwargs):
    service = instance._service_model.metadata["endpointPrefix"]

    return await xray_recorder.record_subsegment_async(
        wrapped, instance, args, kwargs,
        name=service,
        namespace='aws',
        meta_processor=aws_meta_processor,
    )
```
When wrapt runs, the wrapped service is a coroutine which needs to be awaited.
Until its awaited, the injecting of headers wont run.

I created a copy of `record_subsegment` called `record_subsegment_async` which is also a coroutine and it runs  `return_value = await wrapped(*args, **kwargs)` which during that the headers are injected and then the result is returned. After that everything else works.

I tried awaiting the wrapped function inside of `_xray_traced_aiobotocore` but when the headers are injected the subsegment hasn't begun as thats done inside of `record_subsegment`.

What would you reckon is the best way for solving this, I was thinking of subclassing xray_recorder but if there is a simpler way I'm up for that. Then another PR will come your way :smile:.

terrycain:
Closes #5 

Added Async functions for begin subsegment and capture decorator.

Also fixed typo from my last PR

The `yield from` syntax is not valid syntax in py27 so was causing issues when I tried to add the async functions in the recorder module. Even tried some evil generator hacks but then that didn't work when py3 tried to use the functions.
Instead the recorder is subclassed to provide some py3 compatible async functions. When the SDK is imported in py3, the async compatible recorder class will be chosen.

The `get_new_stubbed_recorder` function was also updated to match the above logic. All the test cases were modified to explicitly set the context, as I was having issues where botocore was using an AsyncContext and then getting in a mess.

Updated the docs to mention if one is using aioboto3/aiobotocore that they should use the AsyncContext and provided a snippet.

Tox seems to be running ok on py27,34,35,36.


tests/test_async_recorder.py
```diff
@@ -0,0 +1,21 @@
...
+
+xray_recorder = get_new_stubbed_recorder()
+
+
+@xray_recorder.capture_async('test_1')
+async def async_method():
+    pass
+
+
+async def test_capture(loop):
```
haotianw465: Could you add a test for nested async functions? An decorated async function calls another decorated async function should result in two nested subsegments.

tox.ini
```diff
@@ -17,9 +17,10 @@ deps =
     # Python3.5+ only deps
     py{35,36}: aiohttp >= 2.3.0
     py{35,36}: pytest-aiohttp
+    py{35,36}: aiobotocore
 
 commands =
-    py{27,34}: coverage run --source aws_xray_sdk -m py.test tests --ignore tests/ext/aiohttp --ignore tests/test_async_local_storage.py
+    py{27,34}: coverage run --source aws_xray_sdk -m py.test tests --ignore tests/ext/aiohttp --ignore tests/ext/aiobotocore --ignore tests/test_async_local_storage.py --ignore tests/test_async_recorder.py
```
haotianw465: Again the keyword you used for capture_async actually works with 3.4 but the aiobotocore patcher only work with 3.5.

aws_xray_sdk/ext/aiobotocore/patch.py
```diff
@@ -0,0 +1,164 @@
...
+    )
+
+    wrapt.wrap_function_wrapper(
+        'aiobotocore.endpoint',
+        'AioEndpoint._encode_headers',
+        _inject_header,
+    )
+
+
+async def _xray_traced_aiobotocore(wrapped, instance, args, kwargs):
```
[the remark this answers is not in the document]
terrycain: Ahh, my bad. It was a sort of lazy copy and paste, originally the methods in the async recorder were in the main recorder which had @coroutine on them from compat. Currently them being @asyncio.coroutine will allow them to be used in py34 without complaints, aiobotocore uses aiohttp which has dropped py34 support and is now using `async def` syntax. So either the async recorder stays compatible with py34 and people can use it with other frameworks, or we just draw a line and just do async purely on py35+. Your choice, I have no preference

aws_xray_sdk/ext/aiobotocore/patch.py
```diff
@@ -0,0 +1,164 @@
...
+    return result
+
+
+def _inject_header(wrapped, instance, args, kwargs):
+    headers = args[0]
+    inject_trace_header(headers, xray_recorder.current_subsegment())
+    return wrapped(*args, **kwargs)
+
+
+def aws_meta_processor(wrapped, instance, args, kwargs,
```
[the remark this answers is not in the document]
terrycain: Yep, good shout, Most of the functions in that module can also be shared.

tests/test_async_recorder.py
terrycain: Ahh cool I didnt realise that, but it makes sense. Will do

tox.ini
terrycain: Have used async/await throughout so best to exclude those tests from py34

terrycain:
I also updated the code that loads in json to use pkg_resources so then the module can also be used in egg/zip form without being extracted.

Had to do a `from __future__ import absolute_import` in `ext/boto_utils.py` as importing the `botocore.exceptions` module was failing in py27 as botocore folder was in the same directory.

aws_xray_sdk/ext/aiobotocore/patch.py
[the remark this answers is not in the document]
haotianw465: Yes we should support py35+ only.

tox.ini
haotianw465: Yeah I mean let's get all async related stuff to be 3.5+ only.

terrycain: Yeah :smile:, async in 3.4 is pretty nasty anyway. I "think" I've excluded 3.4 from anything async and purged any `@asyncio.coroutine`'s too

haotianw465:
One last thing is that I like the new structure of ext/resources/botocore_para_whitelist.json. Could we rename it to aws_para_whitelist.json? It's more obvious as more and more files added under resources/ folder. Also botocore causes some confusion on if that json file is effective to boto3 or aiobotocore etc.

terrycain:
Sure, yeah exactly.

terrycain:
Should be good

haotianw465:
Oh one thing I forgot is that feel free to add `capture_async` and aiohttp middleware support to the repo landing page. To be more specific the [README.md](https://github.com/aws/aws-xray-sdk-python/blob/master/README.md) `Quick Start` section.

terrycain:
Sure, I'll push some more changes tomorrow.

terrycain:
Added capture async example and a slimmed down example of adding middleware to aiohttp

PR #8, opened 2017-12-01
Removing Entity Name invalid characters

levesquejf:
If you create subsegments with invalid Name characters, this causes issues in the X-Ray traces. This Pull Request remove the invalid characters and adds a warning in the log. The valid character list has been found on http://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html.


aws_xray_sdk/core/models/entity.py
```diff
@@ -12,21 +13,26 @@
 
 
 log = logging.getLogger(__name__)
+_valid_name_characters = string.ascii_letters + string.digits + '_.:/%&#=+\-@'
```
haotianw465: It would be useful to add the doc link here as a comment for future reference.

aws_xray_sdk/core/models/entity.py
```diff
@@ -12,21 +13,26 @@
...
 
         # required attributes
         self.id = self._generate_random_id()
-        self.name = name
+        self.name = ''.join([c for c in name if c in _valid_name_characters])
         self.start_time = time.time()
         self.parent_id = None
 
+        if self.name != name:
+            log.warning("Removing Name invalid characters.")
```
haotianw465: Use segment/subsegment name instead of name could be clearer when user see such log entries.

PR #10, opened 2017-12-04
Fixing Django 2.0 and TestClass tests errors

levesquejf:
By adding __test__ = False in the TestServer Class, pytest ignores it so we don't get the warning message.

Django 2.0 is out and it causes errors with actual tests. I never used Django so I am not the best person to port the tests to 2.0 but with this change in tox.ini, Python 3 tests are not failing.

### DIFF
--- a/tests/ext/aiohttp/test_aiohttp.py
+++ b/tests/ext/aiohttp/test_aiohttp.py
@@ -19,6 +19,7 @@ class CustomStubbedEmitter(UDPEmitter):
     """
     Custom stubbed emitter which stores all segments instead of the last one
     """
+
     def __init__(self, daemon_address='127.0.0.1:2000'):
         super(CustomStubbedEmitter, self).__init__(daemon_address)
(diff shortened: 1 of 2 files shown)


haotianw465:
I think this is fine. Since 2.0 is a new major version, we need to do some proper testing before claiming support for Django 2.0. One thing I want to confirm is that do you mean to prevent py2 test from failing?Django 2.0 drops py2 support but running with py3+ should be fine.

levesquejf:
The modules are not the same in Django 1.x and 2.0 so the tests on py3 are failing. It is also failing in py2 (unless you keep an old version in .tox as I did, my mistake). By removing .tox directory, without my changes, all versions are failing.

adamchainz:
The only reason Django bumped to 2.0 from 1.0 was dropping Python 3 support, the level of change is otherwise no more/less than other Django version changes (which aren't small however, reading the release notes is a good idea).

PR #11, opened 2017-12-07
Fix error code flag when using Python 3

Dunedan:
When using Python 3 the error code flag was never set, thanks to the
division changes between Python 2 and Python 3.

The behavior before this commit was:

Python2:

```
>>> status_code = 403
>>> status_code / 100 == 4
True
```

Python3:

```
>>> status_code = 403
>>> status_code / 100 == 4
False
```


Dunedan:
Ah, right. That explains why it wasn't caught earlier. I pushed a new commit to remove the flagging from `boto_utils` completely.

haotianw465:
Hi, this looks good to me. Could you merge your two local commits into one so I can merge that single commit to the master?

Dunedan:
Here you go. :smile:

PR #13, opened 2017-12-08
Support patching PynamoDB calls to DynamoDB

Dunedan:
While [PynamoDB][1] uses `botocore` under the hood it doesn't call
`botocore`'s [`BaseClient._make_api_call`][2]. Instead it implements its own
[`Connection._make_api_call`][2], making it necessary to separately patch
PynamoDB, even though it uses `botocore`.

As PynamoDB's `Connection._make_api_call` doesn't return information
about the response as `botocore` does, there is no response information
to pass to AWS X-Ray.

For adding the trace header patching `botocore` is sufficient, as
PynamoDB does use `botocore`'s prepared requests. To make that work out
of the box the overall patching logic got adjusted to automatically
patch `botocore` as well when pynamodb patching is selected.

For tests I tried to do something useful, but only came up with the
single test which calls real DynamoDB.

I'm not sure if patching `_make_api_call` is the right approach at all
or if it would be better to patch the
`requests.session.Session.send`-call in PynamoDB. That would give
access to the response data, while causing a separate subsegment per
retry in case of throttling.

So I'm looking forward to get some feedback.

All in all it's pretty annoying that PynamoDB needs additional logic
while it uses `botocore` and `requests` which are both already
supported.

[1]: https://github.com/pynamodb/PynamoDB
[2]: https://github.com/boto/botocore/blob/06f34bb4409da12efb20f154e943d04d47cfe748/botocore/client.py#L568-L614
[3]: https://github.com/pynamodb/PynamoDB/blob/d100b6594ade5381192cfd42874ac2588a02fd1f/pynamodb/connection/base.py#L335-L418

### DIFF
--- a/aws_xray_sdk/ext/pynamodb/patch.py
+++ b/aws_xray_sdk/ext/pynamodb/patch.py
@@ -1,0 +1,63 @@
+import botocore.vendored.requests.sessions
+import json
+import wrapt
+
+from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.core.models import http
+from aws_xray_sdk.ext.boto_utils import _extract_whitelisted_params
+
+
+def patch():
+    """Patch PynamoDB so it generates subsegements when calling DynamoDB."""
+    if hasattr(botocore.vendored.requests.sessions, '_xray_enabled'):
+        return
+    setattr(botocore.vendored.requests.sessions, '_xray_enabled', True)
+
+    wrapt.wrap_function_wrapper(
+        'botocore.vendored.requests.sessions',
+        'Session.send',
+        _xray_traced_pynamodb,
+    )
+
+
+def _xray_traced_pynamodb(wrapped, instance, args, kwargs):
+
+    # Check if it's a request to DynamoDB and return otherwise.
+    try:
+        service = args[0].headers['X-Amz-Target'].decode('utf-8').split('_')[0]
+    except KeyError:
+        return wrapped(*args, **kwargs)
+    if service.lower() != 'dynamodb':
+        return wrapped(*args, **kwargs)
+
+    return xray_recorder.record_subsegment(
+        wrapped, instance, args, kwargs,
+        name='dynamodb',
+        namespace='aws',
+        meta_processor=pynamodb_meta_processor,
+    )
+
+
+def pynamodb_meta_processor(wrapped, instance, args, kwargs, return_value,
+                            exception, subsegment, stack):
+    operation_name = args[0].headers['X-Amz-Target'].decode('utf-8').split('.')[1]
+    region = args[0].url.split('.')[1]
+    request_id = return_value.headers.get('x-amzn-RequestId')
+
+    aws_meta = {
+        'operation': operation_name,
+        'request_id': request_id,
+        'region': region
+    }
+
+    if exception:
+        subsegment.add_error_flag()
+        subsegment.add_exception(exception, stack, True)
+
+    subsegment.put_http_meta(http.STATUS, return_value.status_code)
+
+    _extract_whitelisted_params(subsegment.name, operation_name,
+                                aws_meta, [None, json.loads(args[0].body)],
+                                None, return_value.json())
+
+    subsegment.set_aws(aws_meta)
--- a/docs/thirdparty.rst
+++ b/docs/thirdparty.rst
@@ -6,10 +6,11 @@ Third Party Library Support
 Patching Supported Libraries
 ----------------------------
 
-The SDK supports aioboto3, aiobotocore, boto3, botocore, requests, sqlite3 and mysql-connector.
+The SDK supports aioboto3, aiobotocore, boto3, botocore, pynamodb, requests, sqlite3 and
+mysql-connector.
 
 To patch, use code like the following in the main app::
-    
+
     from aws_xray_sdk.core import patch_all
 
     patch_all()
@@ -30,12 +31,16 @@ The following modules are availble to patch::
         'aiobotocore',
         'boto3',
         'botocore',
+        'pynamodb',
         'requests',
         'sqlite3',
         'mysql',
     )
 
-Patching boto3 and botocore are equivalent since boto3 depends on botocore
+Patching boto3 and botocore are equivalent since boto3 depends on botocore.
+
+Patching pynamodb applies the botocore patch as well, as it uses the logic from the botocore
+patch to apply the trace header.
 
 Patching mysql
 ----------------------------
--- a/tests/ext/pynamodb/test_pynamodb.py
+++ b/tests/ext/pynamodb/test_pynamodb.py
@@ -1,0 +1,74 @@
+import botocore.session
+import pytest
+from botocore.exceptions import ClientError
+from pynamodb.attributes import UnicodeAttribute
+from pynamodb.exceptions import VerboseClientError
+from pynamodb.models import Model
+
+from aws_xray_sdk.core import patch
+from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.core.context import Context
+
+patch(('pynamodb',))
+
+
+@pytest.fixture(autouse=True)
+def construct_ctx():
+    """
+    Clean up context storage on each test run and begin a segment
+    so that later subsegment can be attached. After each test run
+    it cleans up context storage again.
+    """
+    xray_recorder.configure(service='test', sampling=False, context=Context())
+    xray_recorder.clear_trace_entities()
+    xray_recorder.begin_segment('name')
+    yield
+    xray_recorder.clear_trace_entities()
+
+
+def test_exception():
+    class SampleModel(Model):
+        class Meta:
+            region = 'us-west-2'
+            table_name = 'mytable'
+
+        sample_attribute = UnicodeAttribute(hash_key=True)
+
+    try:
+        SampleModel.describe_table()
+    except VerboseClientError:
+        pass
+
+    subsegments = xray_recorder.current_segment().subsegments
+    assert len(subsegments) == 1
+    subsegment = subsegments[0]
+    assert subsegment.name == 'dynamodb'
+    assert len(subsegment.subsegments) == 0
+    assert subsegment.error
+
+    aws_meta = subsegment.aws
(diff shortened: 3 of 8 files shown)


haotianw465:
Good call. Patching `requests.session.Session.send` works better in this case. There are three things you probably most care about regarding talking to DynamoDB: 1: network roundtrip time, 2: return status code 3: internal retries.

For current patching you only get 1. But patching the botocore vendored requests you get all three benefits. Though it means a little bit more work on parameter capture since you go one level deeper on the code path. 

The X-Ray SDK support requests only if your application has direct dependency on it. It leaves botocore vendored requests untouched. So when you patch PynamoDB you can directly dive into the vendored requests and do the necessary patch there. You just need to make sure when you have for example 2 retries, it creates 3 sibling subsegments (1 original call + 2 retries) where for each subsegment it represents a DynamoDB call. This way you gain more value on tracing the DynamoDB calls from your contribution.

The AWS Go SDK has hooks that provide insights on internal retries and TLS info for each retry. Check what you can get from X-Ray Go SDK https://github.com/aws/aws-xray-sdk-go. Although boto doesn't provide as much as Go you can still get more insights on why a call is taking long.

Dunedan:
> You just need to make sure when you have for example 2 retries, it creates 3 sibling subsegments (1 original call + 2 retries) where for each subsegment it represents a DynamoDB call.

Is there a way to mark retries as retries or is it just about creating a separate subsegment per retry?

haotianw465:
It will just create a separate subsegment per retry. This make sure you get the timing and status code for every single retry for more insights. A retry count is not good enough IMO. And as a general library support it'd be better to give more insights to fit more use cases.

Dunedan:
I just pushed a new commit, now directly patching `botocore.vendored.requests.sessions.Session.send`. That does now also log response status and parameters and should create one subsegment per retry as well.

I'd appreciate if you could take another look at it.

docs/thirdparty.rst
```diff
@@ -27,12 +28,15 @@ The following modules are availble to patch::
...
+        'pynamodb',
         'requests',
         'sqlite3',
         'mysql',
     )
 
-Patching boto3 and botocore are equivalent since boto3 depends on botocore
+Patching boto3 and botocore are equivalent since boto3 depends on botocore.
+
+Patching pynamodb patches botocore as well, as it depends partially on botocore.
```
haotianw465: I would suggest put more details on what is patched for `pynamodb` so other users utilizing your work have more details and know the implication of patching `pynamodb`.

tests/ext/pynamodb/test_pynamodb.py
```diff
@@ -0,0 +1,47 @@
...
+        sample_attribute = UnicodeAttribute(hash_key=True)
+
+    try:
+        SampleModel.describe_table()
+    except VerboseClientError:
+        pass
+
+    subsegment = xray_recorder.current_segment().subsegments[0]
+    assert subsegment.error
+    assert subsegment.http['response']['status'] == 400
```
haotianw465: What exception are you trying to produce? Is the code always guaranteed to be 400 or it could be different depend on users and machines?

aws_xray_sdk/ext/pynamodb/patch.py
```diff
@@ -0,0 +1,61 @@
...
+        wrapped, instance, args, kwargs,
+        name='dynamodb',
+        namespace='aws',
+        meta_processor=pynamodb_meta_processor,
+    )
+
+
+def pynamodb_meta_processor(wrapped, instance, args, kwargs, return_value,
+                            exception, subsegment, stack):
+    operation_name = args[0].headers['X-Amz-Target'].decode('utf-8').split('.')[1]
```
haotianw465: Is it possible to capture `request_id` here?

tests/ext/pynamodb/test_pynamodb.py
```diff
@@ -0,0 +1,47 @@
...
+    it cleans up context storage again.
+    """
+    xray_recorder.configure(service='test', sampling=False, context=Context())
+    xray_recorder.clear_trace_entities()
+    xray_recorder.begin_segment('name')
+    yield
+    xray_recorder.clear_trace_entities()
+
+
+def test_exception():
```
haotianw465: Could you add a test to guard the behavior where vendored requests only emit subsegments when talking to DynamoDB? In other words after patching `pynamodb` other AWS services are still captured only through original patched botocore code.

haotianw465:
I fixed some formatting issue on docs/thirdparty.rst when doing SDK 0.95 release and looks like that caused some conflicts. Sorry for the inconvenience. Could you pull the latest commits, resolve the conflicts and merge your commits into one single commit? Thanks.

aws_xray_sdk/ext/pynamodb/patch.py
Dunedan: Thanks for pointing out. Turned out I searched for it in the request instead of the response. Added it now.

docs/thirdparty.rst
Dunedan: I find it pretty difficult to write proper tests for the whole stuff, thanks to all the patching. I added a test now to check that only a single subsegment gets created when the pynamodb patch is enabled, but another AWS service is called. That single subsegment gets created, because the botocore patch is active in this case too.

tests/ext/pynamodb/test_pynamodb.py
Dunedan: I find it pretty difficult to write proper tests for the whole stuff, thanks to all the patching. I added a test now to check that only a single subsegment gets created when the pynamodb patch is enabled, but another AWS service is called. That single subsegment gets created, because the botocore patch is active in this case too.

tests/ext/pynamodb/test_pynamodb.py
Dunedan: These tests currently do real requests to DynamoDB, as I don't know how to avoid them. Using the botocore stubber, as in the botocore tests, doesn't work, because we're patching requests here directly. The requests made by the tests currently return a access denied unless there are is access to resources called "mytable" and "mybucket" available.

docs/thirdparty.rst
Dunedan: I extended the paragraph a bit. I hope that's sufficient.

Dunedan:
All remarks should be addressed now. Waiting for the next round of feedback. :smile:

docs/thirdparty.rst
```diff
@@ -30,12 +31,18 @@ The following modules are availble to patch::
...
         'sqlite3',
         'mysql',
     )
 
-Patching boto3 and botocore are equivalent since boto3 depends on botocore
+Patching boto3 and botocore are equivalent since boto3 depends on botocore.
+
+Patching pynamodb applies the botocore patch as well, as it uses the logic from the botocore
+patch to apply the trace header. As capturing the pynamodb calls relies on intercepting all
+calls made via botocore separately from the botocore patch, there is a small performance penalty
```
haotianw465: I would not mention the `performance penalty` here. An extra if condition is really neglectable IMO considering how much code is actually executed in botocore.

tests/ext/pynamodb/test_pynamodb.py
```diff
@@ -0,0 +1,73 @@
...
+        SampleModel.describe_table()
+    except VerboseClientError:
+        pass
+
+    subsegments = xray_recorder.current_segment().subsegments
+    assert len(subsegments) == 1
+    subsegment = subsegments[0]
+    assert len(subsegment.subsegments) == 0
+    assert subsegment.error
+    assert subsegment.http['response']['status'] in [400, 403]
```
haotianw465: I would suggest remove status code assertion. Any 4xx could be returned depend on setup of the developer's machine.

tests/ext/pynamodb/test_pynamodb.py
```diff
@@ -0,0 +1,73 @@
...
+    As the pynamodb patch applies the botocore patch as well, we need
+    to ensure that only one subsegment is created for all calls not
+    made by PynamoDB. As PynamoDB calls botocore differently than the
+    botocore patch expects we also just get a single subsegment per
+    PynamoDB call.
+    """
+    session = botocore.session.get_session()
+    s3 = session.create_client('s3', region_name='us-west-2')
+    try:
+        s3.get_bucket_location(Bucket='mybucket')
```
haotianw465: You're making external http calls anyway on the dynamo test case so I would suggest to not cause a Client error here. Exception raised from remote endpoint makes sure it reach further down to the botocore code path.

tests/ext/pynamodb/test_pynamodb.py
```diff
@@ -0,0 +1,73 @@
...
+    except ClientError:
+        pass
+
+    subsegments = xray_recorder.current_segment().subsegments
+    assert len(subsegments) == 1
+    assert len(subsegments[0].subsegments) == 0
```
haotianw465: A little bit more assertion to make sure this subsegment is not some random subsegment generated anywhere else. e.g. name assertion should be enough.

tests/ext/pynamodb/test_pynamodb.py
Dunedan: I'm afraid I don't get your point. I am doing an external call in this test as well and as the remote end returns an error (because I'm not authorized to access that bucket) I get an error back. I have to catch this error to be able to properly do the assertions afterwards. Am I missing something?

docs/thirdparty.rst
Dunedan: done

tests/ext/pynamodb/test_pynamodb.py
Dunedan: done

tests/ext/pynamodb/test_pynamodb.py
Dunedan: done

tests/ext/pynamodb/test_pynamodb.py
haotianw465: My bad. I got confused by `ClientError`. I thought it was raised by library due to invalid parameters. In that case there is not any http related code executed.

PR #14, opened 2017-12-10
Add Tracing for SQLAlchemy and Flask-SQLAlcemy

therealryanbonham:
Traces calls from SQLAlchemy ORM query and session classes, and calls from Flask-SQLAlchemy.
Also patches tox.ini so that py36 is used for coverage run, as if your default python was 27 it errors.

Added a couple of basic unit test for both sqlalchemy and flask-sqlalcemy tracing. Updated README as well.   

Welcome feedback, this is my first attempt to extend X-Ray Tracing.


aws_xray_sdk/ext/sqlalchemy/query.py
```diff
@@ -0,0 +1,61 @@
...
+            xray_recorder.end_subsegment()
+        return res
+    return wrapper
+    
+@decorate_all_functions(xray_on_call)   
+class XRaySession(Session):
+    pass
+
+@decorate_all_functions(xray_on_call)   
+class XRayQuery(Query):
```
haotianw465: Any reason you decorate functions from Query class?

aws_xray_sdk/ext/sqlalchemy/query.py
```diff
@@ -0,0 +1,61 @@
+from __future__ import print_function
```
haotianw465: Unused?

setup.py
```diff
@@ -35,7 +35,7 @@
         'Programming Language :: Python :: 3.6',
     ],
 
-    install_requires=['jsonpickle', 'wrapt', 'requests'],
+    install_requires=['jsonpickle', 'wrapt', 'requests', 'future'],
```
haotianw465: Do you need future if you remove all print statement?

README.md
```diff
@@ -175,7 +175,61 @@ app.router.add_get("/", handler)
 
 web.run_app(app)
 ```
+**Add Flask middleware**
```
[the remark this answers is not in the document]
therealryanbonham: Yes. Removing.

README.md
```diff
@@ -175,7 +175,61 @@ app.router.add_get("/", handler)
...
+
+
+**Use SQLAlchemy ORM**
+The SQLAlchemy integration requires you to override the Session and Query Classes for SQL Alchemy
+
+SQLAlchemy integration uses subsegments so you need to have a segment started before you make a query.
+
+```python
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy import *
```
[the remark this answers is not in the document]
therealryanbonham: No, in fact that entire import can go. I refactored the way i was doing imports and that is not needed.

README.md
```diff
@@ -175,7 +175,61 @@ app.router.add_get("/", handler)
...
+from sqlalchemy.orm import sessionmaker
+from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.core.context import Context
+from aws_xray_sdk.ext.sqlalchemy.query import XRaySessionMaker
+
+xray_recorder.configure(service='test', sampling=False, context=Context())
```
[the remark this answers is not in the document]
therealryanbonham: Done

README.md
```diff
@@ -175,7 +175,61 @@ app.router.add_get("/", handler)
...
+xray_recorder.clear_trace_entities()
+xray_recorder.begin_segment('SQLAlchemyTest')
+
+Session = XRaySessionMaker(bind=engine)
+session = Session()
 
+xray_recorder.end_segment()
+app = Flask(__name__)
```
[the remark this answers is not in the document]
therealryanbonham: You have to initialize the app from Flask in order to Initialize Flask-SQLAlchemy

db = XRayFlaskSqlAlchemy(app)

 was just trying to show complete example..

aws_xray_sdk/ext/flask_sqlalchemy/query.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import print_function
```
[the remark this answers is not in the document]
therealryanbonham: Done. Was used for debugging.

aws_xray_sdk/ext/flask_sqlalchemy/query.py
```diff
@@ -0,0 +1,95 @@
...
+from aws_xray_sdk.core.context import Context
+from sqlalchemy.orm.base import _generative
+from sqlalchemy.orm.query import Query
+from flask_sqlalchemy.model import Model
+from sqlalchemy.orm.session import Session, sessionmaker
+from functools import wraps
+from flask_sqlalchemy import SQLAlchemy, BaseQuery, _SessionSignalEvents, get_state
+from aws_xray_sdk.ext.sqlalchemy.query import XRaySession, XRayQuery
+
+def decorate_all_functions(function_decorator):
```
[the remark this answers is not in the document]
therealryanbonham: Flask-SQLAlchemy is a separate python package then Flask, which is why i kept them separated.  Willing to change if needed.

aws_xray_sdk/ext/flask_sqlalchemy/query.py
```diff
@@ -0,0 +1,95 @@
...
+        if getattr(c._local, 'entities', None) is  not None:
+             trace = xray_recorder.begin_subsegment(class_name+'.'+func.__name__)
+        else:
+            trace = None
+        res = func(*args, **kw)
+        if trace is not None:
+            if class_name == 'sqlalchemy.orm.query':
+                for arg in args:
+                    if isinstance(arg, aws_xray_sdk.ext.sqlalchemy.query.XRayQuery):
+                        trace.put_metadata("sql", str(arg));
```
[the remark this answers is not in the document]
therealryanbonham: So just change this call to set_sql(). Is there any special keys for this sql metadata, or can I just the key sql as I have have here?

aws_xray_sdk/ext/sqlalchemy/query.py
therealryanbonham: Query Class called by flask-sqlalchemyt, so this gets us complete traces

aws_xray_sdk/ext/sqlalchemy/query.py
therealryanbonham: Removed

aws_xray_sdk/ext/sqlalchemy/query.py
```diff
@@ -0,0 +1,61 @@
...
+import aws_xray_sdk
+from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.core.context import Context
+from sqlalchemy.orm.base import _generative
+from sqlalchemy.orm.query import Query
+from sqlalchemy.orm.session import Session, sessionmaker
+from functools import wraps
+
+
+def decorate_all_functions(function_decorator):
```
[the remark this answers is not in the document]
therealryanbonham: Yes good point. I meant to go back and do this, and completely forgot.

setup.py
therealryanbonham: No future will still be needed for the super() builtin
from builtins import super

README.md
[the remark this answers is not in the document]
haotianw465: Sounds reasonable. I'm fine with this.

aws_xray_sdk/ext/flask_sqlalchemy/query.py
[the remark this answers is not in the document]
haotianw465: I think we can leave it for now.

haotianw465:
Hi, could you take a look at http://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-sql
When you are sending data to X-Ray back-end service, if you have invalid key under subsegment.sql, your data will not be rejected, but the data will not propagate through our system. Some of the listed keys in that linked doc might be unavailable for specific sql libraries, which is OK. 

Also for the actual query recording you have to submit that change to a separate CR. We have more strict policies for reviewing code that related to query recording. And also the query is required to be sanitized.

tests/ext/flask_sqlalchemy/test_query.py
```diff
@@ -0,0 +1,82 @@
...
+class User(db.Model):
+    __tablename__ = "users"
+
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(255), nullable=False, unique=True)
+    fullname = db.Column(db.String(255), nullable=False)
+    password = db.Column(db.String(255), nullable=False)
+
+
+def _search_entity(entity, name):
```
haotianw465: This along with find_sub could also be a helper class shared between your two query_test.py.  And since these two functions are very general they could be under tests/ root so any other unit test can utilize your contribution.

tests/ext/flask_sqlalchemy/test_query.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            if "subsegments" in entity:
+                for s in entity['subsegments']:
+                    result = _search_entity(s, name)
+                    if result is not None:
+                        return result
+    return None
+
+
+def find_sub(segment, name):
```
haotianw465: I would recommend rename it to find_subsegment. Abbreviate on functions is not a good practice in general if it is not absolutely needed.

therealryanbonham:
@haotianw465  I made the change and used the set_sql to set the sanitized SQL, and added one unit test to test sanitization. That should not me an issue with sqlalchemy orm, as it passes everything via params and those are not included in the sql output i exposed. I will go ahead and comment out the set_sql and test for that, and then make second PR for that work once this first batch is accepted.

haotianw465:
Hi, per doc link http://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html#api-segmentdocuments-sql
This is the sample sql section
```
  "sql" : {
    "url": "jdbc:postgresql://aawijb5u25wdoy.cpamxznpdoq8.us-west-2.rds.amazonaws.com:5432/ebdb",
    "preparation": "statement",
    "database_type": "PostgreSQL",
    "database_version": "9.5.4",
    "driver_version": "PostgreSQL 9.4.1211.jre7",
    "user" : "dbuser",
    "sanitized_query" : "SELECT  *  FROM  customers  WHERE  customer_id=?;"
  }
```
Could you try to get some other info and add them to the sql section as well? For an architecture that has multiple databases the info like `database_type`, `user`, `url` could be very helpful for customers. 

For sql subsegment name we have convention of using databas@host just like the one in the sample showed `"name": "ebdb@aawijb5u25wdoy.cpamxznpdoq8.us-west-2.rds.amazonaws.com"`. This makes sure when you have multiple databases the back-end won't aggregate the latency altogether.

therealryanbonham:
I have a separate branch started to get the set_sql working. I have it adding: url, user, database_type and sanitized_query.

https://github.com/therealryanbonham/aws-xray-sdk-python/tree/set_sql

One issue i see with the way this tracing I implemented works is it is tracing each function in sqlalchemy and thus setting a local namespace on the subsegments, so it is not creating a remove segment.. 

Thoughts? Want that as a seperate PR?

haotianw465:
`trace = xray_recorder.begin_subsegment(class_name+'.'+func.__name__)` could you change the variable name from `trace` to `subsegments`? Trace is a collection of segments and subsegments. Then you can do `subsegment.namespace = 'remote'`. Database subsegment should always be remote.

For setting sql metadata could you have `url` `user` `database_type` included in this PR and a separated PR only for `sanitized_query`?

therealryanbonham:
The combination or using the class and function name as the subsegment name, and namespace of remote, makes a very noisy service map, as every remote call shows up as it's own object. It seems remote namespace wants the name of the subsegment to be something like the hostname you are calling.

Something like 
xray_recorder.begin_subsegment(sql['url'])
instead of 
xray_recorder.begin_subsegment(class_name+'.'+func.__name__)

The issue in that though is the trace become just a bunch of repeat named subsegment calls.. It seems like remote namespaces need a way to be named, separate from the subsegment name...  so that multiple sql functions can be traced as subsegments to a single remote namespace.. 

Something like 

xray_recorder.begin_subsegment(class_name+'.'+func.__name__, namespace=remote, namespace_name=sql['url'])

Is that supported in any way? I have tried to figure out some solution to this but am coming up blank...

haotianw465:
Hi, as I pointed out in the previous thread:
For sql subsegment name we have convention of using databas@host just like the one in the sample showed `"name": "ebdb@aawijb5u25wdoy.cpamxznpdoq8.us-west-2.rds.amazonaws.com"`. This provides proper level of aggregation on service graph. 

Using url only will result in all databases on that instance aggregated to one node, and looks like the naming of your current PR result in too many nodes on service graph. 

The mentioned naming convention works well on all of our SDKs so far and we haven't heard any complain. But it really depends on if the patching code can retrieve the info about database name and endpoint. In this case looks like you can have all the info you need to construct `database@endpoint`.

haotianw465:
Hi, the PR is very close to a ready state. Do you have any update on SQL subsegment naming? Or do you prefer us to take whatever you have right now and finish the work by ourselves?

therealryanbonham:
@haotianw465 I think this is ready for review again. It has SQL URL, user, and datatype added to the sql metadata. The sanatized_query is commented out for this PR. Traces are named by URL and have the function name as an annotation with the key "sqlalchemy".   

I added another set of helper function to util.py to search for subsegments based on annotation key/value and updated unit test to use this as well.

aws_xray_sdk/ext/sqlalchemy/util/decerators.py
```diff
@@ -0,0 +1,101 @@
...
+    m = re.match(r"Engine\((.*?)\)", str(bind))
+    if m is not None:
+        u = urlparse(m.group(1))
+        # Add Scheme to uses_netloc or // will be missing from url.
+        uses_netloc.append(u.scheme)
+        safe_url = ""
+        if u.password is None:
+            safe_url = u.geturl()
+        else:
+            # String password from URL
```
haotianw465: Do you mean strip?

aws_xray_sdk/ext/sqlalchemy/util/decerators.py
```diff
@@ -0,0 +1,101 @@
...
+                    try:
+                        sql = parse_bind(arg.session.bind)
+                        # Commented our for later PR
+                        # sql['sanitized_query'] = str(arg)
+                    except:
+                        sql = None
+        if sql is not None:
+            if getattr(c._local, 'entities', None) is not None:
+                subsegment = xray_recorder.begin_subsegment(sql['url'], namespace='remote')
+                # subsegment = xray_recorder.begin_subsegment(class_name+'.'+func.__name__, )
```
haotianw465: Nitpick. Probably should be removed?

tests/ext/sqlalchemy/test_query.py
```diff
@@ -0,0 +1,70 @@
...
+    subsegment = find_subsegment_by_annotation(xray_recorder.current_segment(), 'sqlalchemy', 'sqlalchemy.orm.session.add')
+    assert subsegment['annotations']['sqlalchemy'] == 'sqlalchemy.orm.session.add'
+    assert subsegment['sql']['url']
+
+
+def test_filter(capsys, session):
+    """ Test calling all() on get all records.
+    Verify we run the query and return the SQL as metdata"""
+    # with capsys.disabled():
+    with capsys.disabled():
```
haotianw465: Duplicate code on line63 and 64?

PR #22, opened 2018-02-17
Filter invalid characters from annotation key names; raise warning if found

### LINKED ISSUE #18: Annotations with special characters in name are silently ignored
Consider this code:

    s = xray_recorder.begin_subsegment('test')
    s.put_annotation('log.mydata', 42)

It is executed without any error or exception, but the annotation just does not appear in XRay console.
If I look at the raw segment data I will see `"annotations": {}` - empty object for annotations. Looks like it is either an issue with XRay backend (silently ignores values with name containing special characters like dot `.` or colon `:`) or an issue with docs & this framework (docs should mention this limitation, and the framework should prohibit such names).

tay-bird:
Fixes #18 

I took the approach of stripping invalid characters from annotation key names and raising a warning, rather than exiting the function.

### DIFF
--- a/aws_xray_sdk/core/models/entity.py
+++ b/aws_xray_sdk/core/models/entity.py
@@ -16,6 +16,7 @@ log = logging.getLogger(__name__)
 
 # List of valid characters found at http://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html
 _valid_name_characters = string.ascii_letters + string.digits + '_.:/%&#=+\-@ '
+_valid_annotation_key_characters = string.ascii_letters + string.digits + '_'
 
 
 class Entity(object):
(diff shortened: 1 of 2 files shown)


aws_xray_sdk/core/models/entity.py
```diff
@@ -140,7 +141,12 @@ def put_annotation(self, key, value):
             log.warning("ignoring unsupported annotation value type %s.", type(value))
             return
 
-        self.annotations[key] = value
+        sanitized_key = ''.join([c for c in key if c in _valid_annotation_key_characters])
```
haotianw465: Annotation key  is difference to entity name. User can have 100 lines of code adding annotations to a single entity. Under worst case this could result in 100 times of string concatenation, which adds a lot of overhead. 

Invalid key is a user error and the key should be dropped until the author explicit fix it using valid characters.

tay-bird:
Thank you for the review, I will make revisions shortly.

tay-bird:
@haotianw465 revisions made.

haotianw465:
It looks all good. Sorry that I forgot to mention if you could add two line check on this unit test https://github.com/aws/aws-xray-sdk-python/blob/master/tests/test_trace_entities.py#L47 to cover the new key dropping logic. Thanks.

tay-bird:
Sure thing.

tay-bird:
@haotianw465 test updated!

PR #24, opened 2018-02-23
httplib fixes

thehesiod:
adds unpatching


thehesiod:
updated changelog, unsure about api doc ref as the other methods in the util module don't seem to be documented under docs/

haotianw465:
My bad. It is here https://github.com/aws/aws-xray-sdk-python/blob/master/docs/aws_xray_sdk.ext.rst but no need to update as it is not a new module.

Issue #28, opened 2018-02-27
Aiohttp improvements, feedback wanted


pfreixes:
Hi folks,

I'm thinking in work on a pair of PR to improve the functionality of AWS-XRay for the Aiohttp framework.

First of all, I would like to add a new `TraceConfig` object - the new Aiohttp Client tracing system since 3.0 [1] - that will allow users just make something like that to inject the specific headers:

```python
import aiohttp

from aws_xray_sdk.ext.aiohttp.client import TraceConfig

async with aiohttp.ClientSession(trace_configs=[TraceConfig()]) as client:
    await client.get('http://example.com/some/redirect/')
```

My main concern with that is how to deal with different Aiohttp versions, is there any requirement or advice with that? 

Second, I would like to deprecate - remove? - the current middleware implementation that uses a deprecated pattern since the 2.X version, we strongly recommend the new one  [2], any concerns?.

[1] https://docs.aiohttp.org/en/stable/client_advanced.html#client-tracing
[2] https://docs.aiohttp.org/en/stable/web_advanced.html?highlight=middleware#middlewares

haotianw465:
Hi, in general the current functionality the SDK already supports for 2.x should remain functioning. Since 3.0 is released on this month, I would assume it takes some time for most of the aiohttp users to upgrade to 3.x.

But to be more specific, support of `aiohttp` includes two parts: middleware and client. The SDK right now only provides a `aiohttp` middleware but it doesn't provide patch code for the client. So you can implement a whole 3.x support with new middleware, client/config, or you can just do the client/config and stick with the 2.x middleware for a while.

For `TraceConfig` I think the naming is too general if your intention is just to carry enough information to inject trace header to the outbound http call. Could you check module [TraceHeader](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/trace_header.py) to see if it fits your needs? 

Also for client support it involves two things: one is to inject trace header to the call, the other is to capture this call as a subsegment. You can check the `requests` patcher to see how a call made via `requests` is captured and at the meantime has trace header injected: https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/requests/patch.py

Please let me know your thoughts and I'm happy to review your PR and merge it as soon as possible.

haotianw465:
Close this since I believe we reached an agreement for doing client support only for 3.x for now.

PR #29, opened 2018-03-01
Use the official middleware pattern for Aiohttp ext

pfreixes:
The previous implementation was using the former version of Aiohttp that
got deprecated since 2.3.0. Related to #28.

The interface didn't change, so from the point of the view of the user, it does not suppose a breaking change. I would suggest using a minor release.


haotianw465:
I have one question. Based on https://aiohttp.readthedocs.io/en/stable/web_advanced.html#aiohttp-web-middlewares: "Prior to v2.3 middleware required an outer middleware factory which returned the middleware coroutine."

Does it mean this PR drops support for `aiohttp` pre-2.3 versions? What happens if one consumer is using 2.2 and upgrade the X-Ray SDK to 0.97 with this change?

pfreixes:
What is expected is people having installed at leasr the last version of 2 major release, they wont have breaking changes and will be "free" of known bugs.

If somebody is still in an older release  the update to the last 2.X should be friction less.

In any case the 3.X version is getting mature, 3.0.5 is out and sooner than later a 3.1 would be released. What IMHO might be the good candidate for production sysy
tems.

haotianw465:
I totally understand what you are saying. The issue is not about whether they upgrade to 2.3 or 3.x, it is about when. We cannot assume upgrading `aiohttp` is at top priorities for all of our customers who depend on `aiohttp`. And a minor release with this also breaks `semver` convention and the expectations on minor upgrade.

We definitely want the `aiohttp` support to use the best practices. I purpose to copy the current middleware to another class and make necessary changes, and then note the old one is deprecated. In the user guide and `README` we can advertising only the new class.

This way whoever new to X-Ray and have `aiohttp` 2.3+ can always get the latest and greatest, and existing X-Ray users running older `aiohttp` can upgrade X-Ray SDK freely. 

Please let me know your thoughts.

pfreixes:
Looking into the middleware code looks like that there is a dependency - the `remote` attrribute [1] - that was introduced in the 2.3.0 series. So, having this dependency I guess that the backward compatibility does not make sense.

Thoughts ?


[1] https://github.com/aio-libs/aiohttp/commit/44edff8a3a930e4ecc35773a48c300c29b9919ae

haotianw465:
You are right. It looks like the first PR for `aiohttp` has implicit requirement on 2.3+. 

Could you add new section "unreleased" on `CHANGELOG` and put this change there? This PR is the first change after 0.96. Then I will go a head and merge it in.

Also it would be helpful if you can add a ">2.3" at https://github.com/aws/aws-xray-sdk-python/blob/master/docs/index.rst "aiohttp item"

pfreixes:
Done. In any case Im working also with the Client part that is only compatible with Aiohttp 3.X series, but lets discuss that in the next MR.

Issue #31, opened 2018-03-05
Removing Segment/Subsugment Name invalid characters.


harveyslash:
When using Mysql with flask sqlalchemy, i get this log: 
```Removing Segment/Subsugment Name invalid characters.```

My SQLALCHEMY URI IS : 

mysql://user:pass@url.something.ap-south-1.rds.amazonaws.com:3306/semiprod?charset=utf8mb4

Even if I remove the ?charset, it seems to be automatically added by sqlalchemy

haotianw465:
Hi, could you share the final subsegment name representing your SQL transaction? We will take a look soon.

harveyslash:
@therealryanbonham's solves it

haotianw465:
Close in favor of https://github.com/aws/aws-xray-sdk-python/issues/36

PR #32, opened 2018-03-05
Improve SubSegment Naming 

therealryanbonham:
Remove Invalid Chars from URL used in name to prevent errors being logged.

### DIFF
--- a/aws_xray_sdk/ext/sqlalchemy/util/decerators.py
+++ b/aws_xray_sdk/ext/sqlalchemy/util/decerators.py
@@ -1,5 +1,6 @@
 import re
 from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.ext.util import strip_url
 from aws_xray_sdk.core.models.entity import _valid_name_characters
 from future.standard_library import install_aliases
 install_aliases()
@@ -46,7 +47,10 @@ def xray_on_call(cls, func):
                         sql = None
         if sql is not None:
             if getattr(c._local, 'entities', None) is not None:
-                sub_name  = ''.join([c for c in sql['url'] if c in _valid_name_characters])
+                # Strip URL of ? and following text
+                sub_name = strip_url(sql['url'])
+                # Ensure url has a valid characters. 
+                sub_name = ''.join([c for c in sub_name if c in _valid_name_characters])
                 subsegment = xray_recorder.begin_subsegment(sub_name, namespace='remote')
             else:
                 subsegment = None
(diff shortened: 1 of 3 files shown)


therealryanbonham:
Addresses https://github.com/aws/aws-xray-sdk-python/issues/31

haotianw465:
Thank you for providing the fix. The log warning is at `entity` level to indicate the actual name is not as same as the original intent whether it is user specified or patcher code. Having the log warning usually means a bug in user code where unsupported characters are used. But in this case it reflects an issue when the SQL Alchemy patcher is assembling the name. 

IMO selectively copy the logic from `entity` class can be a workaround. But we cannot merge this as a real fix. The subsegment name should not contain "?" or anything after "?" at the first place, so the fix should correct the information extraction from the connection string.

thehesiod:
btw I added this for httplib: https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/util.py#L97  may prove useful

therealryanbonham:
@haotianw465  This uses strip_url now as well as doing the logic check from entity. I also added the segment/subsegment name to the log message about invalid chars to make debuging easier for people.

aws_xray_sdk/ext/sqlalchemy/util/decerators.py
```diff
@@ -46,7 +47,11 @@ def wrapper(*args, **kw):
                         sql = None
         if sql is not None:
             if getattr(c._local, 'entities', None) is not None:
-                subsegment = xray_recorder.begin_subsegment(sql['url'], namespace='remote')
+                # Strip URL of ? and following text
+                sub_name = strip_url(sql['url'])
+                # Ensure url has a valid characters. 
+                sub_name = ''.join([c for c in sub_name if c in _valid_name_characters])
```
haotianw465: This is a duplicate to https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/entity.py#L33. IMO when we see a warning on invalid characters we fix from its source, like the `strip_url` here. But we should not deliberately skip the log warning by placing ` sub_name = ''.join([c for c in sub_name if c in _valid_name_characters])` everywhere. 

Otherwise it looks fine.

therealryanbonham: Without that code, we would still end up logging warnings for certain SQL URLs. For example 'échéanciers' is a valid name or a database in postgres, but would be stripped to 'chanciers' for x-ray... I left this code  to prevent the warnings  as the sqlalchemy plugin is setting this, so it is not in users control.. If you want it removed I will, let me know.

haotianw465: I'm fine with special casing the sql subsegment name. I will create a follow up issue for further write an utility module that strips unicode for patchers that might capture unicode as subsegment name.

haotianw465:
Oh one thing I forgot is that could you add one line on `CHANGELOG` "unreleased" section?

haotianw465:
@thehesiod it's nice to have for a utility class to convert some certain unicode characters to ASCII letters. I will make sure to include that on the follow up ISSUE.

PR #34, opened 2018-03-07
Sqlalchemy query

therealryanbonham:
Add SQL Query to X-Ray Metadata for SQLAlchemy and Flask-SQLAlchemy

### DIFF
--- a/aws_xray_sdk/ext/sqlalchemy/util/decerators.py
+++ b/aws_xray_sdk/ext/sqlalchemy/util/decerators.py
@@ -1,8 +1,11 @@
 import re
+import logging
 from aws_xray_sdk.core import xray_recorder
 from future.standard_library import install_aliases
 install_aliases()
 from urllib.parse import urlparse, uses_netloc
+
+log = logging.getLogger(__name__)
 
 
 def decorate_all_functions(function_decorator):
@@ -81,7 +84,7 @@ def parse_bind(bind):
     m = re.match(r"Engine\((.*?)\)", str(bind))
     if m is not None:
         u = urlparse(m.group(1))
-        print('URL: {}'.format(u.get_url()))
+        log.debug('URL: {}'.format(u.get_url()))
         # Add Scheme to uses_netloc or // will be missing from url.
         uses_netloc.append(u.scheme)
         safe_url = ""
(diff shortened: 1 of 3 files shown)


haotianw465:
Thank you for the contribution and we will get our security team review it as soon as possible. At the meantime could you explain a little bit of why `sql['sanitized_query'] = str(arg)` is sufficient? It seems like this one doesn't remove any user data.

therealryanbonham:
I just updated the unit test to better/correctly show why this is valid.

The str(arg) only returns a paramaratized query. and will not contain data values.. so 

session.query(User).filter(User.password=="mypassword!").first()

produces

SELECT users.id AS users_id, users.name AS users_name, users.fullname AS users_fullname, users.password AS users_password
FROM users
WHERE users.password = ?

haotianw465:
Thanks for the explanation.

haotianw465:
Hi, we are in the process of doing security review for this PR. I have one question regarding the SQL expression capture. 

I read through http://docs.sqlalchemy.org/en/latest/orm/tutorial.html#querying but it doesn't mention SQLAlchemy intentionally separate the SQL expression and  user specified parameters into two args under the hood.

The unit test covered the `filter` operation but there are multiple ones like `text`, `join` and `delete` etc. If you do `str(arg)` when they are captured could it contain any sensitive information?

Some extra unit test for other operations or official docs mentioning about how SQL expression is generated would be great.

shouichi:
```
statement = query.statement
print(statement.compile(someengine))
```

> The above forms will render the SQL statement as it is passed to the Python DBAPI, which includes that bound parameters are not rendered inline. SQLAlchemy normally does not stringify bound parameters, as this is handled appropriately by the Python DBAPI

http://docs.sqlalchemy.org/en/latest/faq/sqlexpressions.html

According to the official doc, it seems like it's safe to `str(arg)`. If you want to be extra care, I can take this over and add more test cases.

Thanks!

haotianw465:
Thank you for point out the official SQLAlchemy doc. I believe this is enough and I will merge it in once all conflicts are resolved.

haotianw465:
Manually merged due to some conflicts in GA release.

PR #38, opened 2018-03-14
Recorder.configure: don't overwrite plugins unless specified

MarSoft:
With most options, Recorder.configure() doesn't touch the value if corresponding option is not passed.
But with `plugins` option behaviour is different: it is updated every time you call `configure`.
So this code won't work correctly:

    xray_recorder.configure(plugins=('EC2Plugin', 'ECS'))
    ...
    xray_recorder.configure(context=MyContext())  # this resets plugins

This patch will fix that behaviour: now in order to explicitly reset plugins, you will need to pass an empty tuple to `configure`;
if `plugins is None` then plugin list won't be changed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,8 @@ unreleased
 ==========
 * feature: Use the official middleware pattern for Aiohttp ext. `PR29 <https://github.com/aws/aws-xray-sdk-python/pull/29>`_.
 * bugfix: SQLAlcemy plugin would cause warning messages with some db connection strings that contained invalid characters for a segment/subsegment name.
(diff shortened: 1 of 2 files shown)


haotianw465:
What do you thing of doing this:

```python
         if plugins:
             plugin_modules = get_plugin_modules(plugins)
             for module in plugin_modules:
                 module.initialize()
             self._plugins = plugin_modules
```

MarSoft:
This approach makes it impossible to remove previously added plugins.

MarSoft:
I.e. it will only change plugins list if at least one plugin is specified, with no option to pass empty list (which will be just ignored). Hence I added "is None" check.

haotianw465:
OK make sense.

haotianw465:
Could you also add a line on `CHANGELOG` under "unreleased" section.

MarSoft:
Done

PR #39, opened 2018-03-14
Handle Middleware closing segment before we end all subsegments in sqlalchemy.

therealryanbonham:
In some cases the Flask Middleware closes the segment before we call sqlalchemy.orm.session.flush which results in an AlreadyEndedException exception. This patch catches the exception, logs a debug message and moves on, allowing the Flask app to finish processing.


haotianw465:
Hi, the reason we have this `AlreadyEndedException ` exception is because sometimes users have race condition or bugs result in ending the segment before attaching a new subsegment. 

You can still close the segment while leave subsegments opening, but if a subsegment is created after segment close, it probably is not considered part of the logical operation.

In this case, if Flask already close the segment, it means the request is already served and a response is returned. The session flush happens after response seems should not be part of a request/response cycle so this operation probably should not be captured, instead of randomly being dropped due to `AlreadyEndedException `.

But my question is that should Flask even close the segment when it doesn't know flush succeed or not? What happen if a request mutates some data and gets a 200 but back-end database write fails? I'm little confused.

To make sure we get the right fix I would like to first understand the use case of how recording flush would be useful in this case.

haotianw465:
Hi @therealryanbonham,

Any update on this?

haotianw465:
Closing this for now. Feel free to comment for re-open or create a new one.

PR #41, opened 2018-03-15
Fix invalid aiohttp import

pfreixes:
A silly bug was introduced because of #29, for an unknown reason the test suit didn't detect that issue.

The `web` submodule is not automatically imported by `aiohttp` when we make the typical `import aiohttp`, we have to import it explicitly.

### DIFF
--- a/aws_xray_sdk/ext/aiohttp/middleware.py
+++ b/aws_xray_sdk/ext/aiohttp/middleware.py
@@ -1,15 +1,15 @@
 """
 AioHttp Middleware
 """
-import aiohttp
 import traceback
+from aiohttp import web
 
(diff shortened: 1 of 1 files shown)


haotianw465:
Just want to confirm, did you find this bug when you actually instrumenting `aiohttp` App? With this fix does `tox` still pass?

pfreixes:
Yes, raised when I was instrumenting an Aiohttp app with the `master` branch, more about this discussion here [1], looks like that is intentional so we should fix it.

I've just used `pytest` for that for the `aiohttp` tests, and they are passing - as before.


[1] https://github.com/aio-libs/aiohttp/pull/2838

PR #42, opened 2018-03-19
Support fot Aiohttp Client tracing, aiohttp> 3.0.0

pfreixes:
New `aws_xray_trace_config()` function to retrieve a `aiohttp.TraceConfig` object ready to be used in any `ClientSession`, once the Aiohttp Client session is instantiated with this trace config all sampled HTTP calls will be traced as subsegments and the data properly uploaded.

Usage example:

```python

from aws_xray_sdk.ext.aiohttp.client import aws_xray_trace_config

trace_config = aws_xray_trace_config()
async with ClientSession(loop=loop, trace_configs=[trace_config]) as session:
     async with session.get(url) as resp
        await resp.read()
```


aws_xray_sdk/ext/aiohttp/client.py
```diff
@@ -0,0 +1,53 @@
...
+async def end_subsegment_with_exception(session, trace_config_ctx, params):
+    subsegment = xray_recorder.current_subsegment()
+    subsegment.add_exception(
+        params.exception,
+        traceback.extract_stack(limit=xray_recorder._max_trace_back)
+    )
+    xray_recorder.end_subsegment()
+
+
+def aws_xray_trace_config(name=None, namespace=None):
```
haotianw465: Third party library support should have non-None default value as the library has the context of what is captured. For example, `boto` defaults the namespace to `aws` and `requests` defaults the namespace to `remote`. The aiohttp client is used for making outbound calls so it should be default to `remote` with the option to be set to something else (although I cannot think of any use case that anyone needs to do that).

Setting namespace to `remote` makes sure the service graph renders a node for that endpoint. The X-Ray backend only renders a node when 1. that node actively sends segment or 2. the namespace of the subsegment representing that node is `aws` or `remote`. To be more specific, if you use the client to call a service `B` which you don't own, then `B` is unlikely to generate any trace data. If this call is marked as `remote`, you can see `B` on your service graph.

The only case where you might want to mark an out-going http call subsegment to `local` depends on the http client implementation. If the client thrown a `ClientError` when the call cannot even go out due to connection timeout, DNS lookup fails, invalid input parameters etc, you might want to flip it back to `local` as the error has nothing to do with the downstream service. This is most time a bonus as a lot of libraries doesn't let you easily categorize if the error is from local or from downstream.

tox.ini
```diff
@@ -19,17 +20,26 @@ deps =
...
 
 commands =
     py{27,34}: coverage run --source aws_xray_sdk -m py.test tests --ignore tests/ext/aiohttp --ignore tests/ext/aiobotocore --ignore tests/test_async_local_storage.py --ignore tests/test_async_recorder.py
-    py{35,36}: coverage run --source aws_xray_sdk -m py.test tests
+    py{35,36}: coverage run --source aws_xray_sdk -m py.test tests --ignore tests/ext/aiohttp/test_client.py
 
 setenv =
     DJANGO_SETTINGS_MODULE = tests.ext.django.app.settings
 
+[testenv:py36-aiohttp3]
```
haotianw465: Does this mean the client 3.x support requires Py3.6+?

tests/ext/aiohttp/test_client.py
```diff
@@ -0,0 +1,131 @@
...
+                pass
+        except Exception:
+            # prevent uncatch exception from breaking test run
+            pass
+
+    subsegment = xray_recorder.current_segment().subsegments[0]
+    assert subsegment.fault
+
+    exception = subsegment.cause['exceptions'][0]
+    assert exception.type == 'ClientConnectorError'
```
haotianw465: This is where a `local` namespace makes sense. But it is optional to optimize `end_subsegment_with_exception` as it is not realistic to categorize every single error thrown by the library.

aws_xray_sdk/ext/aiohttp/client.py
pfreixes: Good to know, so better remove it as a param having it as a constant using always remote?

tests/ext/aiohttp/test_client.py
pfreixes: Nops but tox uses 3.5.2 that is not compatible, aiohttp needs at least 3.5.3. Any suggestion?

tox.ini
pfreixes: Nops but tox uses 3.5.2 that is not compatible, aiohttp needs at least 3.5.3. Any suggestion?

tests/ext/aiohttp/test_client.py
pfreixes: Forget about last comment. Its kinda weir that the place where the error comes from is handled by the value of the namespace. 

We can try to mark the most known exceptions as local and others as remote. In any case my doubt here is that taking into accout that the namespace is used to create the subsegment and the error is catched after the creation, can I modify the namespqce once is created?

haotianw465: It's the other way around I believe: the namespace might need to be flipped back to `remote` depending on where the error comes from.

In this case the default captured subsegments should be `remote` as previously discussed. You can flip it back to `local` especially an error like this is thrown. After all non such endpoint could possibly exist so there won't be any downstream.

aws_xray_sdk/ext/aiohttp/client.py
haotianw465: I would purpose to change to `def aws_xray_trace_config(name=None, namespace='remote'):`.

tox.ini
haotianw465: Not sure if I understand this correctly. Tox will use pre-installed interpreter and skip non-exist ones. If your system installs py3.5.2 it will pick that up. If your system doesn't have py3.5 installed it will just skip py35 testing. I'd prefer to add py35 here so it is not mistakenly considered that this new feature only works with py3.6+.

haotianw465:
To more thing:  one is the change log 'unreleased' section needs to be updated. The other one is the `quick start` on `README` has content about adding `aiohttp middleware`. It would be good for the title changed to `working with aiohttp` and update the content to cover both middleware and client.

pfreixes:
@haotianw465 all change that you asked applied, please take a look.

Regarding the `tox` issue I've added a new environment for the Python 3.5 that basically replicates the Python 3.6, I'm sure that there is an alternative way to make so that should follow the tox principles, but I didn't manage to make it. So any advice to improve the current implementation will be welcomed.

haotianw465:
It all looks good. Just one question left.

There is `skip_missing_interpreters = True` in `tox.ini` file. What happens if you merge py35 and py36 in one environment? Does it fail on your end? If your system doesn't have py35 installed or have >=3.5.3 installed it should be fine I believe.

pfreixes:
I didn't merge it because I didn't find a way to make it in a clean way not because I wanted to the tox passed for my environment. To sum up, I'm a newbie with tox and after hitting my head on the same wall many times I gave up. So, if you have a better proposal for the tox .. I will be really glad to hear you :)

haotianw465:
This is not a blocking issue. We will potentially do some tweaks to `tox.ini` to make it better.

pfreixes:
@haotianw465 any chance to have a new release? I think that there are some interesting features in master that users will appreciate it, among them us.

haotianw465:
Yes. I cannot provide a timeline but I am definitely trying to get the new changes out as soon as possible.

haotianw465:
@pfreixes just letting you know 0.97 with all latest changes has been released.

pfreixes:
many thanks !

PR #44, opened 2018-03-21
Do not swallow return_value

### LINKED ISSUE #43: Swallowed `return_value` in `record_subsegment(...)`?
My Python may not be super strong, but in my reading of `try: except: finally:`, it seems that the `return_value` is being trashed in the `LOG_ERROR` case.

From https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/recorder.py:
```
        return_value = None

        try:
            return_value = wrapped(*args, **kwargs)
            return return_value
        except Exception as e:
            exception = e
            stack = traceback.extract_stack(limit=self._max_trace_back)
            raise
        finally:
            # No-op if subsegment is `None` due to `LOG_ERROR`.
            if subsegment is None:
                return
```

Isn't that last `return` actually equivalent to `return None`?

nelz9999:
Fixes #43


haotianw465:
Thank you for your contribution. It all looks good. Could you also add this PR link to `CHANGELOG` unreleased section? I will merge this in once that change is done.

nelz9999:
@haotianw465 Done!

PR #58, opened 2018-05-10
Give up when there is no open segment and LOG_ERROR is configured

pfreixes:
Without setting this flag as False, any HTTP call done with the Aiohttp
Client when there is no an active segment will fail with the
SegmentNotFoundException exception.

This issue can be annoying for this systems that do not create
explicitly the segements and relay on the sampled incomming requests,
which create automatically segments just for a sub set of the incomming
request.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ CHANGELOG
 unreleased
 ==========
 * feature: Added Sqlalchemy parameterized query capture. `PR34 <https://github.com/aws/aws-xray-sdk-python/pull/34>`_.
+* bugfix: Added new `raise_if_not_subsegment` parameter for Aiohttp Client tracing `PR58 <https://github.com/aws/aws-xray-sdk-python/pull/58>`_.
 
 1.0
 ===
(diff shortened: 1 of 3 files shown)


pfreixes:
Please consider this as a potential bug that might require a minor release.

pfreixes:
Closing, the segment is always created by the middleware ...  If I've found something different I will reopen it. Sorry for the noise

haotianw465:
Hi, did you check `AWS_XRAY_CONTEXT_MISSING` on this page? https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-python-configuration.html

If the aiohttp client errors out due to segment is `None` in case of `LOG_ERROR`, then the aiohttp client patcher code needs to be fixed so it uses the original function when segment is not present.

pfreixes:
You are right. In any case, PR fixed to take into account that Aiohttp client tracing must give up silently when the `context_missing` is configured as `LOG_ERROR`.

So, yes it needs to be fixed yet.

pfreixes:
Any chance to have a minor release?

haotianw465:
Yes we will get a minor release as soon as possible. Thanks for your patience.

PR #59, opened 2018-05-15
Handle Aiohttp Exceptions as valid responses

pfreixes:
This PR fixes a bug that till now handled all exceptions as just 500,
while valid Aiohttp Exceptions have to be handled as valid responses.

They are used to notify none 2XX responses, perhaps a 401 is notified
with an `HTTPUnauthorized` exception.


pfreixes:
Please, considere it in WIP I will add some changes. With the current code the Aiohttp exception is never reraised to the upper middlewares.

pfreixes:
Please, considere it in WIP I will add some changes. With the current code the Aiohttp exception is never reraised to the upper middlewares.  Better close it, I will reopen it in  while

pfreixes:
Without this PR all none 2XX response code are considered 500, so it needs to be fixed. The PR is ready to be reviewed.

Sorry to be late, I've seen that you have already released the 1.1.  but most likely this PR will deserve also a new release.

/cc @haotianw465

pfreixes:
Sorry to bother you @haotianw465 , do you have any ETA for the next release?

Issue #62, opened 2018-05-24
Chalice support?


xplorld:
Currently this project does not patch Chalice. We are using it and hope for a first-party support.

haotianw465:
Hi, thank you for the feedback. Chalice support is in our backlog and we will work with Chalice team to prioritize this feature. Please stay tuned. Please feel free to let us know any use case or functionalities you are specifically looking for.

xplorld:
Ideally, we can patch Chalice by simply calling `patch_all()` and all invocations in `@app.route` methods will present in XRay as a segment, with all good metadata like path and stuff.

rahulakash:
@xplorld Any updates on the chalice support? We would really love to use XRay with our chalice projects. 

We are hitting the same issue mentioned by @jarikujansuu here -  https://github.com/aws/chalice/issues/464#issuecomment-372063990

dmulter:
My use case is covered by my comment in the Chalice GitHub issue:

There's nice middleware support for Flask, would be awesome to see Chalice treated as a first-class citizen the same way.

https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-python-middleware.html#xray-sdk-python-adding-middleware-flask

phillprice:
Is there an update on the progress of this?

chanchiem:
Hey,

This is still something that's in our backlog and we will work with the Chalice team to prioritize this accordingly. Please stay tuned!

futurama92:
Hi @chanchiem! Is there any update for x-ray?

srprash:
Hi @futurama92 ,
Unfortunately we haven't been able to prioritize the Chalice support on our end. There was an [effort](https://github.com/aws/chalice/pull/1159) made to add the X-Ray support to Chalice on the repo which doesn't seem to be active recently. We absolutely welcome any contribution for this feature.
We will post an update as soon as possible. Thanks for your patience!

PR #66, opened 2018-06-01
add support for chalice

xplorld:
As discussed in #62.

Add subsegment and other annotations.
Name: `MyChaliceApp GET /path/to/endpoint`
Will trace all router calls, but will not trace `lambda_function`s and `schedule`s.

### DIFF
--- a/aws_xray_sdk/core/patcher.py
+++ b/aws_xray_sdk/core/patcher.py
@@ -6,6 +6,7 @@ log = logging.getLogger(__name__)
 SUPPORTED_MODULES = (
     'aiobotocore',
     'botocore',
+    'chalice',
     'pynamodb',
     'requests',
     'sqlite3',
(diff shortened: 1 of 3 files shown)


xplorld:
@haotianw465 what do you think?

haotianw465:
Thank you for the contribution. The code looks good in general. Patching `chalice` private API `Chalice._get_view_function_response` for request capturing might not be ideal and this contract could break at anytime.

The ideal integration IMO is to have Chalice integrate X-Ray by having Chalice framework depend on X-Ray SDK and provide a single option just like `TracingConfig` in Lambda function configuration https://docs.aws.amazon.com/lambda/latest/dg/API_TracingConfig.html. But this requires more work on Chalice team and takes time to happen. Having something ready to use for now has great value too. I will chat with Chalice team internally and come back to you soon.

Issue #67, opened 2018-06-07
unparented segment/subsegment resolution


thehesiod:
I'm using aws-xray-sdk to trace an application which uses the google pubsub module.  Unfortunately this module uses various threads which means that I get a lot of `SegmentNotFoundException` exceptions causing my application to fail.

It's going to be impossible for me to track down all the locations where it creates threads and pass the trace entity and call `xray_recorder.set_trace_entity`.

~How about if this SDK provides support for a "root thread" which it will use to resolve the segment/subsegment if the current context does not have one?~

haotianw465:
Hi, sorry for the late response. I saw the `UnparentedContext` in your PR and this could definitely be an option for context management. But before going further to simply add that, I would like to know more about your exception on tracing under pubsub module.

Here are several options:
1. Set context missing strategy to `LOR_ERROR` so `SegmentNotFoundException` won't be thrown. This way you will lose tracing data that doesn't have a context.
2. Like what you did with the `UnparentedContext`. This way each unparented segment carries a different trace id so you won't be able to easily correlates tracing data sent from various thread. Maybe you are doing some extra work in addition to what you actually pasted in the PR comment?
3. We have an item on our backlog for supporting automatic context propagation from main thread to worker threads with `future.ThreadPoolExecutor`. The downside is that this only fits very limited use case and for some libraries like `s3transfer` you have to go very deep to figure out what threading module it is using.

Please let me know in an ideal case how you expect your tracing data to be structured so I can provide further help.

thehesiod:
late?  that was really fast :)

1. Ya that's an existing option, however we lose these traces like you stated
2. That's what I'm trying now, we'll see how messy it gets :)  Basically the google pubsub client library creates a bunch of background helpers which contact the network and I want to get some visibility into what they're doing.  Unfortunately when they run I have no active "root" thread so they're completely isolated without any hierarchy.  In my scenario I'm using the apscheduler, and while my callback is inactive these threads are firing off network calls.
3. That would be nice but there are cases where the workers do work w/o the thread which owns the pool having an active segment (like my case).

I think it's worth discussing what the ideal solution would be, I have no firm proposal.

haotianw465:
Hi,

Please let me know if solution2 works for you. In that case I would imagine your service graph is split into two parts. One represents the request flow starts from your main thread and the other one represents the background work. All background work subsegments will be aggregated to a single node with name like `unparented segment` and you still see errors, outbound calls etc if applicable. 

[...]

Issue #26, opened 2018-02-23
Add ability to run disabled


thehesiod:
datadog tracing (https://github.com/DataDog/dd-trace-py) has the ability to disable tracing via a `enabled` parameter to their `configure` method (https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/tracer.py#L82).  This is really useful so one can control the behavior of tracing via an environment variable for example without having to change code, allowing for local and prod behaviors.  I suggest a similar mechanism for xray

thehesiod:
if I have any free time I'll try working on a PR, however guidance and recommendations are welcome

haotianw465:
We have some improvement plan on this matter internally but right now we cannot provide more details. I will leave this open to gather feedback from broader audience. But a PR is always welcome for a quick improvement. 

Also keep in mind that it doesn't matter whether you disable tracing or not if the application is not the starting point on your microserivces, as any X-Ray integrated service (apps instrumented with X-Ray SDK, AWS Lambda etc) respects the sampling decision passed from upstream. So if you have a full call path all integrated you will just need to change your frontend sampling to `rate=0` and `fixed_target=0` to completely disable X-Ray tracing.

thehesiod:
ah, perhaps if `enabled=False` it just does that?

haotianw465:
Yes you can definitely do that with a PR. That doesn't conflict with any of our internal work. Please stay tuned. Thanks.

jaysonsantos:
Hey there, this feature is quite important especially if you are running tests as TargetPoller and RulePoller's threads will never die if there is no agent running.
The dirty fix i made was a fixture to kill those threads after pytest finishes testing.
```python
@pytest.yield_fixture('session', autouse=True)
def fix_xray_threads():
    # TODO: This should be removed after https://github.com/aws/aws-xray-sdk-python/issues/26 is solved
    yield
    import ctypes
    import threading
    main_thread = threading.main_thread()
    for thread in threading.enumerate():
        if thread.daemon or thread == main_thread:
            continue

        ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_long(thread.ident), ctypes.py_object(SystemExit))

```

thehesiod:
i forget why we didn't merge my "enabled" param support.

jaysonsantos:
[...]

PR #68, opened 2018-06-07
[WIP] Add threading and startup enhancements

thehesiod:
Adds support for threaded root entity support discusssed in: https://github.com/aws/aws-xray-sdk-python/issues/67

and running in disabled mode per: https://github.com/aws/aws-xray-sdk-python/issues/26


thehesiod:
actually this isn't going to work, coming up with a better impl.  This will need to be discussed in detail.

thehesiod:
closing as I see you can implement a custom context.

thehesiod:
if anyone is curious, this is how I solved it, may be worth adding to the lib as a supported mechanism.

```python
class UnparentedContext(Context):
    def end_subsegment(self, end_time=None):
        """ This differs from the base context in that if we're closing the first subsegment, and the parent
        is an unparented segment, we close the parent. """
        subsegment = self.get_trace_entity()
        if self._is_subsegment(subsegment):
            subsegment.close(end_time)
            self._local.entities.pop()
            root_entity = self._local.entities[0]
            if len(self._local.entities) == 1 and root_entity.name == _UNPARENTED_SEGMENT_NAME:
                xray_recorder.end_segment()

            return True
        else:
            log.warning("No subsegment to end.")
            return False

    def handle_context_missing(self):
        """
        Called whenever there is no trace entity to access or mutate.
        """
        xray_recorder.begin_segment(_UNPARENTED_SEGMENT_NAME)
        return self.get_trace_entity()
```

Issue #90, opened 2018-09-06
Psycopg <class 'IndexError'> tuple index out of range


IlyaSukhanov:
```
  File "/var/task/sqlalchemy/engine/default.py", line 412, in connect
    return self.dbapi.connect(*cargs, **cparams)
  File "/var/task/aws_xray_sdk/ext/psycopg2/patch.py", line 20, in _xray_traced_connect
    dbname = kwargs['dbname'] if 'dbname' in kwargs else re.search(r'dbname=(\S+)\b', args[0]).groups()[0]
```

IlyaSukhanov:
I'm presuming the `host` `dbname` `port` `user` extraction logic does not support all the formats of connection strings that psycopg supports.

IlyaSukhanov:
`psycopg2.connect()` Supports 3 variants of operation:
1. `psycopg2.connect("dbname=test user=postgres password=secret")`
2. `psycopg2.connect(dbname="test", user="postgres", password="secret")`
3. `psycopg2.connect()`

Variant 3 is ok if arguments are passed via env-vars:
> Any other connection parameter supported by the client library/server can be passed either in the connection string or as a keyword. The PostgreSQL documentation contains the complete list of the supported parameters. Also note that the same parameters can be passed to the client library using environment variables.

aws-xray-sdk-python does not support variant 3 or mix of variant 2 and 3.

IlyaSukhanov:
It may not be obvious but use of patch_all() or patch of psycopg will cause any code error out and stack trace when opening connection to database. This is quite serious bug.

IlyaSukhanov:
After some more digging around turns out problem was different but solution remained the same.

Psycopg supports passing of database name both as `dbname` and as `database`. SQL Alchemy uses the `database` variety triggering the bug.
https://github.com/psycopg/psycopg2/blob/master/lib/__init__.py#L96

There may still be a way to use method 3 as described in comment above but its not obvious. And psycopg will error out if there are no arguments on connect. https://github.com/psycopg/psycopg2/blob/master/lib/__init__.py#L122 Maybe if you set port via kwarg / dsn but the rest via env var?

revmischa:
Getting this error when database is localhost

Tankanow:
Is this still an issue or was this solved with 180de75?

haotianw465:
@Tankanow this is fixed in the commit you mentioned. We will close this issue once the commit is released (which should be soon).

haotianw465:
[...]

PR #83, opened 2018-08-29
ISSUE-80: Add patch support for psycopg2 

Tankanow:
Issue 80

Add patch support for [psycopg2](https://pypi.org/project/psycopg2) using existing `dbapi2` tracing classes. For now, add `host`, `database`, `user`, and `database_type` to `sql` metadata.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #91, opened 2018-09-06
Support dbname and database kwarg in psycopg2.connect()

IlyaSukhanov:
*Issue #90 

*Description of changes:
~~psycopg2.connect() Supports 3 variants of operation:~~
 1. ~~psycopg2.connect("dbname=test user=postgres password=secret")~~
 2. ~~psycopg2.connect(dbname="test", user="postgres", password="secret")~~
 3. ~~psycopg2.connect()~~

~~Previously we only supported 1 and 2. Now support variant 3 and mix of 2 and 3.~~

~~This change is not entierly backwards compatible previously the metadata used to
contain keys `url` and `user`. These are no longer exposed however the connection
parameters are available as `connection_string`. If this backward incompatibility
is not desirable then change can be made to extract values as they were before
using conn.dsn. However the URL will not be correct as password is mangled. It's
not clear to me if one option is clearly better than the other.~~

### DIFF
--- a/aws_xray_sdk/ext/psycopg2/patch.py
+++ b/aws_xray_sdk/ext/psycopg2/patch.py
@@ -1,5 +1,6 @@
 import re
 import wrapt
+from operator import methodcaller
 
 from aws_xray_sdk.ext.dbapi2 import XRayTracedConn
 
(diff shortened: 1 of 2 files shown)


IlyaSukhanov:
[Looks like the fields do matter](https://github.com/aws/aws-xray-sdk-python/pull/83/files/d3fc58e3bf10becceea814a15d3ed73f88084f15#r213794779). I'll adjust PR to rebuild them of `conn.dsn()`

IlyaSukhanov:
Made change backwards compatible.

Tankanow:
👍 I like this improvement.

haotianw465:
Thank you for the contribution. This improvement looks good to me. I think it is a miss on unit test where some use cases are not covered https://github.com/aws/aws-xray-sdk-python/blob/master/tests/ext/psycopg2/test_psycopg2.py. Do you mind adding more common use cases in the test to prevent regression?

haotianw465:
Thank you for your contribution. I have merged this PR and will prepare a release soon.

PR #111, opened 2018-11-01
Add params and functions to allow recording and streaming SQL queries

hasier:


hasier:
Sorry didn't intend to open this PR here, just testing around in own fork

hasier:
Updated the branch, should be ready for review.

aws_xray_sdk/ext/django/conf.py
```diff
@@ -14,6 +14,7 @@
     'DYNAMIC_NAMING': None,
     'STREAMING_THRESHOLD': None,
     'MAX_TRACE_BACK': None,
+    'STREAM_SQL': False,
```
haotianw465: Sorry for the late response. If by default the sql should captured, should this be set to `True`?

hasier: No worries, thanks for your comment. I meant it is enabled by default as this is the current behaviour for SQLAlchemy, so I wanted to keep it. But for the rest (in this case Django), I have set it off by default, as it is the behaviour previous to this PR. Hope this makes sense.

haotianw465: Hi, I would not consider an enhancement to be concerned.  We have some other enhancements that added more data and haven't seen any issue. 

The SQLAlchemy query capture was also submitted as a PR. In fact the SDK means to have query capture as the default behavior. It's just due to security concerns it has different development and review cycle. I would suggest to keep the query capture behavior consistent, which that behavior is to have parameterized query ready whenever possible. Thoughts?

hasier: Regarding the default behaviour, I agree, I'll change it so that capture is enabled by default.

Regarding the parametrised queries, afaik there is no way to ensure that the incoming query is a parametrised one... DBAPI2 (PEP 249) recommends this behaviour, but then it is up to the implementation to decide how the queries and parameters are formatted.
Is this the concern you have? Would it be better to attempt the patch in a different place? The only one I can think of is to attempt to patch the Django ORM (I haven't tried, so I don't know how feasible it is). I mention the Django ORM as, for example, psycopg2 also allows different Cursor classes to be used with its connections, which might lead to the same situation again.

hasier: I removed the patch for DBAPI2 and just did so for Django. Please let me know if it makes sense now.

PR #112, opened 2018-11-07
Remove the dependency on requests

Dunedan:
This removes the use of `requests` from the core functionality, which
results in `requests` being only used as extension, therefore not being
necessary as dependency anymore.

The motivation for this change is to get rid of all the dependencies of
`requests`, whose size sums up quite a bit, when not requiring to patch
`requests`. It follows the removal of `requests` from `botocore`, which
was done as part of [v1.11.0](https://botocore.amazonaws.com/v1/documentation/api/latest/index.html#upgrading-to-1-11-0).

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license.

### DIFF
--- a/aws_xray_sdk/core/plugins/ec2_plugin.py
+++ b/aws_xray_sdk/core/plugins/ec2_plugin.py
@@ -1,5 +1,7 @@
 import logging
-import requests
+from future.standard_library import install_aliases
+install_aliases()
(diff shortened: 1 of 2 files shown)


haotianw465:
Thank you for your contribution. I have merged this PR and it will be out with our next release.

PR #113, opened 2018-11-13
Recursively patch any given module functions with capture

hasier:
Accept a list of plain existing modules in `patch` that will recursively traverse all the packages inside and patch all available functions and class methods with `@xray_recorder.capture()`.


aws_xray_sdk/ext/django/apps.py
```diff
@@ -38,6 +38,10 @@ def ready(self):
             max_trace_back=settings.MAX_TRACE_BACK,
         )
 
+        if settings.PATCH_MODULES:
```
haotianw465: Sorry for the late response. Other parts look good. Why capture the patching as a segment during startup? This generates dangling traces and doesn't seem to be useful.

hasier: No worries, thanks for your comment. I found it useful for our project, as we were patching in the `settings` file and the result was similar. It is also less of a hassle to add imports.
Still, I agree it does not look especially good. I gave it a thought back then and didn't come with any other solution... Any ideas? Where do you usually find useful to apply the patching in a Django app?

hasier: Oh I see. Yes, now I remember I added that because during startup I was seeing subsegments that failed to have a parent. I guess I did so because some functions decorated with `.capture()` are being called during startup/patching, mainly due to importing different modules (it's a really big project). I don't think it would be a big issue if we would not have a parent for those subsegments, but back then I guess I wanted to try and catch most of them. Would you do that differently?

haotianw465: I can see recursive patching can become an issue when you have some submodules get patched but they are only called during app startup. I'm OK with an additional Django configuration parameter that create a segment that wraps the startup process. But IMO this should be opt-in as some users will not want to see dangling traces that just contains startup, and this is not an issue for users that don't use this feature.

PR #115, opened 2018-12-17
Add patch support for pg8000 (Pure python driver)

ridha:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/ext/psycopg2/test_psycopg2.py
+++ b/tests/ext/psycopg2/test_psycopg2.py
@@ -39,7 +39,7 @@ def test_execute_dsn_kwargs():
         cur = conn.cursor()
         cur.execute(q)
 
-    subsegment = xray_recorder.current_segment().subsegments[-1]
+    subsegment = xray_recorder.current_segment().subsegments[0]
     assert subsegment.name == 'execute'
     sql = subsegment.sql
     assert sql['database_type'] == 'PostgreSQL'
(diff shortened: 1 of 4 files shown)


tests/ext/psycopg2/test_psycopg2.py
```diff
@@ -39,7 +39,7 @@ def test_execute_dsn_kwargs():
         cur = conn.cursor()
         cur.execute(q)
 
-    subsegment = xray_recorder.current_segment().subsegments[0]
+    subsegment = xray_recorder.current_segment().subsegments[-1]
```
haotianw465: Thank you for the contribution. I have one question. This test setup explicitly patches only `psycopg2` in line 12. How does this get impacted? You can refer `httplib` patcher where it has unpatch method so for test teardown it restores the original `httplib` to avoid impacting other libraries depending on it: https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/httplib/patch.py#L168-L178. It's also required that double instrumentation is not the default for customers running `patch_all`. You can check the logic here: https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/patcher.py#L30-L34

ridha: Thanks for noticing! Fixed in the latest revision.

PR #119, opened 2019-01-17
Add ability to enable/disable the SDK (#26)

chanchiem:
Added the ability to enable/disable the SDK upon configuration. This prevents application unit tests that originally relied on segments to be generated to no longer throw SegmentNotFound exceptions if a unit test tests a method that relies on a previously generated Segment.

*Issue #, if available:*
#26 

*Description of changes:*
* Added SDKConfig to hold SDK-wide configuration information. In our case, whether it's enabled/disabled is kept track of. This is used by the Patcher, Recorder, and the Lambda Context to change behaviors if it's disabled. 
* For the case of the Patcher, patching does not patch any methods. 
* For the Recorder, all segment creation automatically creates DummySegments and therefore not send Segments and subsequent Subsegments to the Daemon. Using DummySegments ensures that downstream methods calls on the segment/subsegment entities don't throw exceptions.
* To deal with the _Capture_ decorator, a Dummy parent segment is automatically generated to mimic middleware segment generation.

*Testing*
* Core Patcher module should not import modules when SDK is disabled
* Each specific middleware (Flask, Django, Aiohttp) is tested to ensure that they produce DummySegments and dont send anything to the emitter. 
* Recorder should produce DummySegments and DummySubsegments
* SDKConfig should take Environment Variables over hardcoded configuration.

### DIFF
--- a/aws_xray_sdk/core/lambda_launcher.py
+++ b/aws_xray_sdk/core/lambda_launcher.py
@@ -2,7 +2,7 @@ import os
 import logging
 import threading
 
-import aws_xray_sdk
+from aws_xray_sdk import global_sdk_config
 from .models.facade_segment import FacadeSegment
 from .models.trace_header import TraceHeader
 from .context import Context
@@ -94,7 +94,7 @@ class LambdaContext(Context):
         """
         header_str = os.getenv(LAMBDA_TRACE_HEADER_KEY)
         trace_header = TraceHeader.from_header_str(header_str)
-        if not aws_xray_sdk.global_sdk_config.sdk_enabled():
+        if not global_sdk_config.sdk_enabled():
             trace_header._sampled = False
 
         segment = getattr(self._local, 'segment', None)
@@ -128,7 +128,7 @@ class LambdaContext(Context):
         set by AWS Lambda and initialize storage for subsegments.
         """
         sampled = None
-        if not aws_xray_sdk.global_sdk_config.sdk_enabled():
+        if not global_sdk_config.sdk_enabled():
             # Force subsequent subsegments to be disabled and turned into DummySegments.
             sampled = False
         elif trace_header.sampled == 0:
--- a/aws_xray_sdk/core/lambda_launcher.py
+++ b/aws_xray_sdk/core/lambda_launcher.py
@@ -2,10 +2,11 @@ import os
 import logging
 import threading
 
+from aws_xray_sdk.sdk_config import SDKConfig
 from .models.facade_segment import FacadeSegment
 from .models.trace_header import TraceHeader
+from .models.dummy_entities import DummySubsegment
 from .context import Context
-
 
 log = logging.getLogger(__name__)
 
@@ -73,6 +74,11 @@ class LambdaContext(Context):
         if not self._is_subsegment(current_entity) and current_entity.initializing:
             log.warning("Subsegment %s discarded due to Lambda worker still initializing" % subsegment.name)
             return
+
+        enabled = SDKConfig.sdk_enabled()
+        if not enabled:
+            # For lambda, if the SDK is not enabled, we force the subsegment to be a dummy segment.
+            subsegment = DummySubsegment(current_entity)
 
         current_entity.add_subsegment(subsegment)
         self._local.entities.append(subsegment)
--- a/aws_xray_sdk/core/lambda_launcher.py
+++ b/aws_xray_sdk/core/lambda_launcher.py
@@ -2,10 +2,9 @@ import os
 import logging
 import threading
 
-from aws_xray_sdk.sdk_config import SDKConfig
+import aws_xray_sdk
 from .models.facade_segment import FacadeSegment
 from .models.trace_header import TraceHeader
-from .models.dummy_entities import DummySubsegment
 from .context import Context
 
 log = logging.getLogger(__name__)
@@ -72,13 +71,9 @@ class LambdaContext(Context):
         current_entity = self.get_trace_entity()
 
         if not self._is_subsegment(current_entity) and current_entity.initializing:
-            log.warning("Subsegment %s discarded due to Lambda worker still initializing" % subsegment.name)
+            if sdk_config_module.sdk_enabled():
+                log.warning("Subsegment %s discarded due to Lambda worker still initializing" % subsegment.name)
             return
-
-        enabled = SDKConfig.sdk_enabled()
-        if not enabled:
-            # For lambda, if the SDK is not enabled, we force the subsegment to be a dummy segment.
-            subsegment = DummySubsegment(current_entity)
 
         current_entity.add_subsegment(subsegment)
         self._local.entities.append(subsegment)
@@ -99,6 +94,9 @@ class LambdaContext(Context):
         """
         header_str = os.getenv(LAMBDA_TRACE_HEADER_KEY)
         trace_header = TraceHeader.from_header_str(header_str)
+        if not aws_xray_sdk.global_sdk_config.sdk_enabled():
+            trace_header._sampled = False
+
         segment = getattr(self._local, 'segment', None)
 
         if segment:
@@ -130,7 +128,10 @@ class LambdaContext(Context):
         set by AWS Lambda and initialize storage for subsegments.
         """
         sampled = None
-        if trace_header.sampled == 0:
+        if not aws_xray_sdk.global_sdk_config.sdk_enabled():
+            # Force subsequent subsegments to be disabled and turned into DummySegments.
+            sampled = False
+        elif trace_header.sampled == 0:
             sampled = False
         elif trace_header.sampled == 1:
             sampled = True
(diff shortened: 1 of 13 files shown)


aws_xray_sdk/core/recorder.py
```diff
@@ -77,7 +80,7 @@ def __init__(self):
         if type(self.sampler).__name__ == 'DefaultSampler':
             self.sampler.load_settings(DaemonConfig(), self.context)
 
-    def configure(self, sampling=None, plugins=None,
+    def configure(self, enabled=None, sampling=None, plugins=None,
```
haotianw465: This seems a conflict with the `SDKConfig`. `SDKConfig` should sit on top of recorder config rather than the other way around. If users change `SDKConfig` it should automatically apply to all the key paths including recorder component.

aws_xray_sdk/sdk_config.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @staticmethod
+    def sdk_enabled():
+        """
+        Returns whether the SDK is enabled or not.
+        """
+        return SDKConfig.__SDK_ENABLED
+
+    @staticmethod
+    def set_sdk_enabled(value):
```
haotianw465: If an arbitrary value is passed, we should fallback to `true` and throw a warning says invalid configuration, rather than taking that arbitrary value and rely on python truthiness to determine. (5 means true while [] means false)

aws_xray_sdk/core/lambda_launcher.py
```diff
@@ -74,6 +75,11 @@ def put_subsegment(self, subsegment):
             log.warning("Subsegment %s discarded due to Lambda worker still initializing" % subsegment.name)
```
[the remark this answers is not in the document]
chanchiem: Done

aws_xray_sdk/core/recorder.py
```diff
@@ -134,10 +146,11 @@ class to have your own implementation of the streaming process.
             by auto-capture. Lower this if a single document becomes too large.
         :param bool stream_sql: Whether SQL query texts should be streamed.
 
-        Environment variables AWS_XRAY_DAEMON_ADDRESS, AWS_XRAY_CONTEXT_MISSING
+        Environment variables AWS_XRAY_ENABLED, AWS_XRAY_DAEMON_ADDRESS, AWS_XRAY_CONTEXT_MISSING
```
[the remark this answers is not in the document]
chanchiem: I agree. Changed.

aws_xray_sdk/core/recorder.py
chanchiem: This would then mean that if customers want to hardcode disable their SDK, they would need to do it only from the SDKConfig module. Is this a favorable API? This may be slightly confusing to customers if they think that doing recorder.configure(enabled=False) would mean that other modules such as the LambdaContext would be disabled. What are your thoughts on this?

aws_xray_sdk/sdk_config.py
chanchiem: That's a good point. Added fallback to true and explicit boolean checking.

aws_xray_sdk/core/recorder.py
haotianw465: `SDKConfig` is already a public module which users can import and set values. Those operations right now already take effect regardless of calling `configure` or not, based on the current implementation. So practically it is a global configuration object.

chanchiem: Good point. Added the changes into the new commit

chanchiem:
Second revision added.

Changes:
    - SDKConfig now built on top of recorder. SDK-level configuration should all be done
      through SDKConfig.
    - Renamed the SDK Enable environment variable to AWS_XRAY_SDK_ENABLED
    - Disabling the SDK now causes LambdaContext to set sampling decision of facade segment to False.
      Previously, it would force all subsegments to be generated as dummy subsegments.
    - SDKConfig.set_sdk_enabled(value) method defaults to True and throws an exception
      if value is not of type boolean.

chanchiem:
Third revision added:

Changes:

- Removed enable flag from recorder. Now it checks the global SDK configuration to determine if the SDK is disabled.
- Global SDK configuration no longer modifies the recorder.
- Tests modified to reflect changes.

aws_xray_sdk/core/lambda_launcher.py
```diff
@@ -2,11 +2,11 @@
 import logging
 import threading
 
+import aws_xray_sdk
```
haotianw465: This import is too general. It should be sth like `from aws_xray_sdk import global_sdk_config`

aws_xray_sdk/core/patcher.py
```diff
@@ -7,6 +7,7 @@
 import sys
 import wrapt
 
+import aws_xray_sdk
```
haotianw465: Same here

aws_xray_sdk/core/patcher.py
```diff
@@ -60,6 +61,9 @@ def _is_valid_import(module):
 
 
 def patch(modules_to_patch, raise_errors=True, ignore_module_patterns=None):
+    enabled = aws_xray_sdk.global_sdk_config.sdk_enabled()
+    if not enabled:
```
haotianw465: A debug level log entry is useful to show why patch logic is skipped

aws_xray_sdk/core/recorder.py
```diff
@@ -5,6 +5,7 @@
 import platform
 import time
 
+import aws_xray_sdk
```
haotianw465: Same

aws_xray_sdk/core/recorder.py
```diff
@@ -18,12 +19,13 @@
...
 from .plugins.utils import get_plugin_modules
 from .lambda_launcher import check_in_lambda
-from .exceptions.exceptions import SegmentNameMissingException
+from .exceptions.exceptions import SegmentNameMissingException, SegmentNotFoundException
 from .utils.compat import string_types
 from .utils import stacktrace
 
 log = logging.getLogger(__name__)
 
+XRAY_ENABLED_KEY = 'AWS_XRAY_ENABLED'
```
haotianw465: We discussed changing this to `AWS_XRAY_SDK_ENABLED` in the previous PR comments

chanchiem: Forgot to remove. Never actually used. Removed

aws_xray_sdk/core/recorder.py
chanchiem: Done

aws_xray_sdk/core/patcher.py
chanchiem: Done

aws_xray_sdk/core/patcher.py
chanchiem: Done

chanchiem:
Fourth Revision Added:

Changes:
* Refactored multiple modules; instead of import aws_xray_sdk to get the global_sdk_config, direct import using from... import.
* Removed unused XRAY_ENABLED_KEY variable from recorder.
* Added debug log entry to the patcher to inform the customer that patching is disabled when the SDK is disabled.

aws_xray_sdk/core/sampling/sampler.py
```diff
@@ -9,6 +9,7 @@
 from .target_poller import TargetPoller
 from .connector import ServiceConnector
 from .reservoir import ReservoirDecision
+import aws_xray_sdk
```
haotianw465: should this import be updated?

aws_xray_sdk/sdk_config.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        # Environment Variables take precedence over hardcoded configurations.
+        if cls.XRAY_ENABLED_KEY in os.environ:
+            cls.__SDK_ENABLED = str(os.getenv(cls.XRAY_ENABLED_KEY, 'true')).lower() != 'false'
+        else:
+            if type(value) == bool:
+                cls.__SDK_ENABLED = value
+            else:
+                cls.__SDK_ENABLED = True
+                raise InvalidParameterTypeException(
```
haotianw465: You probably cannot default to `True` if you are throwing an exception here as the program will break. I would suggest a log warning here.

aws_xray_sdk/core/sampling/sampler.py
chanchiem: Ah good catch. I must've missed this one.

aws_xray_sdk/sdk_config.py
chanchiem: Done

chanchiem:
Fifth revision added:

Changes:
- Refactored sampler to use "from ... import ..." for the global sdk config module.
- Global sdk config logs and defaults to true instead of throwing an exception when an invalid parameter is passed into the method.
- Removed unused key in recorder unit test.

PR #125, opened 2019-01-25
Bring aiobotocore support back

yan12125:
*Description of changes:*

aiobotocore 0.10.0+ is compatible with botocore 1.12+ [1]

[1] https://github.com/aio-libs/aiobotocore/releases/tag/0.10.0


aws_xray_sdk/core/patcher.py
```diff
@@ -12,6 +12,7 @@
 log = logging.getLogger(__name__)
 
 SUPPORTED_MODULES = (
+    'aiobotocore',
```
chanchiem: Can you add 'aiobotocore' as an item in the NO_DOUBLE_PATCH tuple as well?

yan12125: Done!

chanchiem:
Tested and confirmed that aiobotocore is now supported in botocore 1.12+. Patching also works

PR #126, opened 2019-02-06
add suport for IGNORE context missing strategy

djcrabhat:
Adds a new IGNORE strategy for context missing errors.

I often found myself turning off the "aws_xray_sdk.core" in my unit test code, or when debugging a problem locally.  I thought being able to set this strategy would let me know of anything else logged from the package, but would let me silence those context missing errors I'm always gonna get when running something designed for lambda locally.

### DIFF
--- a/aws_xray_sdk/core/context.py
+++ b/aws_xray_sdk/core/context.py
@@ -7,7 +7,7 @@ from .exceptions.exceptions import SegmentNotFoundException
 log = logging.getLogger(__name__)
 
 MISSING_SEGMENT_MSG = 'cannot find the current segment/subsegment, please make sure you have a segment open'
-SUPPORTED_CONTEXT_MISSING = ('RUNTIME_ERROR', 'LOG_ERROR')
+SUPPORTED_CONTEXT_MISSING = ('RUNTIME_ERROR', 'LOG_ERROR', 'IGNORE')
 CXT_MISSING_STRATEGY_KEY = 'AWS_XRAY_CONTEXT_MISSING'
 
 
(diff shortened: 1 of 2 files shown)


chanchiem:
Hey, can we get more information about your use case here? It seems like the point of this PR is to ignore the SegmentNotFound exception so that tests function correctly without segments needing be generated for you (in your case Lambda). This seems related to issue #2 and we are currently in the works for this feature.

djcrabhat:
Ya, that's a good point.  This is a much simpler implementation of that goal: my PR still allows all the patching to be done, starts the poller, it doesn't add all this global config module: it just sorta NOOPs the context missing handler.

(also, that python27 fail on Travis seems to be for a missing ipaddress package, and I didn't wanna mess with `install_requires`in a PR that seemed out of the scope of that)

chanchiem:
That's a good point, but that implementation would work for your particular use case. What about in the case where the code expects a segment to be mutated such as through the following code?
```
xray_recorder.begin_segment("SegmentName")
segment = xray_recorder.current_segment()
segment.put_annotation("some", "data")
```

This case would throw an exception. A general solution is required that would still allow original code paths to be executed without needing to tweak any user code. It needs to be an on/off switch that doesn't require any special-casing at the user level.

chanchiem:
Hey, just an update: the ability to disable the SDK has been merged: https://github.com/aws/aws-xray-sdk-python/pull/119

I will update how to use it on the readme. But basically, there are two ways, one is to do it directly through the module: aws_xray_sdk.global_sdk_config

Or through an environment variable by setting the following key to 'false':
AWS_XRAY_SDK_ENABLED

Let me know if this helps your use case

djcrabhat:
Ya, that's a much better solution than my simple "not logging" approach.  Thanks so much!

PR #127, opened 2019-02-08
Add Serverless Framework Support

chanchiem:
Added support for the Serverless Framework. This is often used by applications that utilize AWS Lambda as the compute and API Gateway as the endpoint which invokes AWS Lambda. Specifically, works for Flask and Django deployments as well as through the tool Zappa.

Removes the "FacadeSegments cannot be mutated" exception error and generates a subsegment whenever segments are created.

*Issue #, if available:*
#2

*Description of changes:*
A new "Serverless" context is created to give the abstraction of Segments being the top level entities but is then converted to a subsegment upon transmission to the data plane. These segments are called MimicSegments and their only purposes is to be treated as a Segment all the way until transmission via the Recorder. All MimicSegments are parented by FacadeSegments.

Currently supports Flask and Django as middlewares; this has been confirmed to be natively working with Zappa if the application is running under Flask/Django.

The middlewares Flask and Django automatically use the ServerlessContext when it is present in a Lambda environment.


aws_xray_sdk/core/models/mimic_segment.py
```diff
@@ -0,0 +1,35 @@
...
+    it is converted to a locally-namespaced, subsegment. This is only done during serialization.
+    All Segment-related method calls done on this object are valid.
+
+    Subsegments are automatically created with the namespace "local" to prevent it from appearing
+    as a node on the service graph. For all purposes, the MimicSegment can be interacted as if it's
+    a real segment, meaning that all methods that exist only in a Segment but not a subsegment
+    is available to be used.
+    """
+
+    def __init__(self, facade_segment=None, original_segment=None):
```
haotianw465: If these arguments are all required, don't give them default. The default `None` should only be given when the argument is optional.

aws_xray_sdk/core/models/mimic_segment.py
```diff
@@ -0,0 +1,35 @@
+from .segment import Segment
+from ..exceptions.exceptions import MimicSegmentInvalidException
+
+
+class MimicSegment(Segment):
+    """
+    The MimicSegment is an entity that mimics a segment for the use of the serverless context.
```
haotianw465: `serverless context` is too generic. This is a special workaround with lambda due to lambda integration implementation details. We should be clear on the intention of this "hacky" class type.

aws_xray_sdk/core/models/mimic_segment.py
```diff
@@ -0,0 +1,35 @@
...
+            raise MimicSegmentInvalidException("Invalid MimicSegment construction. "
+                                               "Please put in the original segment and the facade segment.")
+        super(MimicSegment, self).__init__(name=original_segment.name, entityid=original_segment.id,
+                                           traceid=facade_segment.trace_id, parent_id=facade_segment.id,
+                                           sampled=facade_segment.sampled)
+
+    def __getstate__(self):
+        """
+        Used during serialization. We mark the subsegment properties to let the dataplane know
+        that we want the mimic segment to be represented as a subsegment.
```
haotianw465: let the `X-Ray service` know.   I would prefer `transformed` over `represented`

aws_xray_sdk/core/recorder.py
```diff
@@ -243,7 +243,8 @@ def begin_segment(self, name=None, traceid=None,
             self._populate_runtime_context(segment, decision)
 
         self.context.put_segment(segment)
-        return segment
+        current_segment = self.get_trace_entity()
+        return current_segment
```
haotianw465: What is the intention here?

aws_xray_sdk/core/serverless_context.py
```diff
@@ -0,0 +1,131 @@
...
+from .models.mimic_segment import MimicSegment
+from .context import CXT_MISSING_STRATEGY_KEY
+from .lambda_launcher import LambdaContext
+from .context import Context
+
+
+log = logging.getLogger(__name__)
+
+
+class ServerlessContext(LambdaContext):
```
haotianw465: Again this class name is too generic and will not work well in the long term. It's tailored for web framework + lambda user case given the limitation we currently have.

aws_xray_sdk/core/serverless_context.py
```diff
@@ -0,0 +1,131 @@
+import os
+import logging
+
+from .models.facade_segment import FacadeSegment
+from .models.segment import Segment
+from .models.mimic_segment import MimicSegment
+from .context import CXT_MISSING_STRATEGY_KEY
+from .lambda_launcher import LambdaContext
+from .context import Context
```
haotianw465: This probably should not be used. The design is too complicated when you have 3 different context storage on a single module. The end goal is to make web app works on lambda. Feel free to refactor old classes.

aws_xray_sdk/core/serverless_context.py
```diff
@@ -0,0 +1,131 @@
...
+        :return: Entity
+        """
+        # Call to Context.get_trace_entity() returns the latest mimic segment/subsegment if they exist.
+        # Otherwise, returns None through the following way:
+        # No mimic segment/subsegment exists so Context calls LambdaContext's handle_context_missing().
+        # By default, Lambda's method returns no-op, so it will return None to ServerlessContext.
+        # Take that None as an indication to return the rightful handle_context_missing(), otherwise
+        # return the entity.
+        entity = Context.get_trace_entity(self)
+        if entity is None:
```
haotianw465: Interestingly lambda facade segment will always be present even if somewhere in the web app doesn't propagate context properly. Should we attach whatever we have to the facade segment and throw a warning? There is no meaning to force fail user's application when there IS context available.

aws_xray_sdk/core/serverless_context.py
```diff
@@ -0,0 +1,131 @@
...
+    subsegments underneath the lambda-generated segment. This context also
+    ensures that FacadeSegments exist through underlying calls to _refresh_context().
+    """
+    def __init__(self, context_missing='RUNTIME_ERROR'):
+        super(ServerlessContext, self).__init__()
+
+        strategy = os.getenv(CXT_MISSING_STRATEGY_KEY, context_missing)
+        self._context_missing = strategy
+
+    def put_segment(self, segment):
```
haotianw465: My bold guess is that if you just use LambdaContext's put_subsegment on the "mimicSegment" everything will just work. To me the final json doc on a webapp+lambda is just one of many forms of what the SDKs are creating. The webapp layer trace entity is just THE giant local subsegment which is the single child of facade segment and single parent of many other subsegments. You can easily generate such a special form by just having a capture block on the entire lambda handler function and add some http metadata to that. The design can be simplified if you look at the problem we are trying to solve from a different view.

aws_xray_sdk/core/models/mimic_segment.py
chanchiem: Good point

aws_xray_sdk/core/models/mimic_segment.py
chanchiem: Done

aws_xray_sdk/core/models/mimic_segment.py
chanchiem: I agree. It's much more precise.

aws_xray_sdk/core/recorder.py
chanchiem: In the case when the context modifies the segment under the hood and returns a (potentially) new entity, this ensures that if the returned entity is mutated at the higher levels, it would propagate down to the context.

aws_xray_sdk/core/serverless_context.py
chanchiem: Okay, I think ServerlessLambdaContext makes more sense then. Any thoughts?

aws_xray_sdk/core/serverless_context.py
chanchiem: That's a fair point. I wanted to keep the "storage" portion of the context similar to its grandparent class, "Context" in case any changes were to be made with using ThreadLocal storage. Since it does make the design more complicated, I think explicitly using ThreadLocal instead of relying on the grandparent context to handle segment/subsegment storage would be better.

aws_xray_sdk/core/serverless_context.py
chanchiem: The facade segment is stored in threadlocal's 'segment' field, not the 'entities' list. In the lambda context implementation of get_trace_entity(), if no entities exist in the 'entities' list, then it returns threadlocal.segment, which is the facade segment. In this design, the FacadeSegment is never actually revealed to the higher ups and can't be accessed outside of the context itself. This design gives the customer the abstraction that they can BeginSegments, mutate it, and append subsegments onto it, but only if they have created a segment (which becomes a mimic segment)

This handle_context_missing() would be in the case where the higher-level calls didn't generate a segment. When a segment is not generated, we want to give the abstraction of what would normally happen in the grandparent class "Context", so in this case, throw an exception since the MimicSegment was never generated.

aws_xray_sdk/core/serverless_context.py
chanchiem: PutSubsegment call to the Lambda Context would implicitly call parent_segment.increment(), and in this case, Facade Segments will be the parent, an FacadeSegmentMutationException would be thrown.

chanchiem:
Revision 2 Added:

* Renamed ServerlessContext to ServerlessLambdaContext to be more precise about its purpose.
* Instead of explicitly using Context class's put_subsegment/put_segment methods, call LambdaContext's grandparent class to store segments for code conciseness. LambdaContext's methods are really only used to acquire the FacadeSegment.
* Parameters when initializing the mimic segment are now required.
* Unit tests added to test these changes.

aws_xray_sdk/core/serverless_lambda_context.py
```diff
@@ -0,0 +1,130 @@
...
+from .models.segment import Segment
+from .models.mimic_segment import MimicSegment
+from .context import CXT_MISSING_STRATEGY_KEY
+from .lambda_launcher import LambdaContext
+
+
+log = logging.getLogger(__name__)
+
+
+class ServerlessLambdaContext(LambdaContext):
```
chrisradek: I'm a little unclear on why this additional Context is needed. It looks like `LambdaContext` does a no-op and logs a warning when working with `segments`. Since this is an additive change, it seems that the two Context classes can be merged into one. Is there a backwards compatibility issue this is solving?

chanchiem: That is a good point that you make. I did consider just adding the existing features to the LambdaContext, but I didn't want to confuse customers and suddenly allow them to create "segments" in Lambda apart from using it in a serverless application.

aws_xray_sdk/core/models/mimic_segment.py
```diff
@@ -0,0 +1,39 @@
...
+        if not issubclass(type(original_segment), Segment) or type(facade_segment) != FacadeSegment:
+            raise MimicSegmentInvalidException("Invalid MimicSegment construction. "
+                                               "Please put in the original segment and the facade segment.")
+        super(MimicSegment, self).__init__(name=original_segment.name, entityid=original_segment.id,
+                                           traceid=facade_segment.trace_id, parent_id=facade_segment.id,
+                                           sampled=facade_segment.sampled)
+
+    def __getstate__(self):
+        """
+        Used during serialization. We mark the subsegment properties to let the dataplane know
```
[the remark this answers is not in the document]
chanchiem: Done

aws_xray_sdk/core/serverless_lambda_context.py
```diff
@@ -0,0 +1,130 @@
...
+        entity = super(LambdaContext, self).get_trace_entity()
+        if entity is None:
+            return super(LambdaContext, self).handle_context_missing()
+        else:
+            return entity
+
+    def set_trace_entity(self, trace_entity):
+        """
+        Store the input trace_entity to local context. It will overwrite all
+        existing ones if there is any.
```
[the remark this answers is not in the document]
chanchiem: Done

aws_xray_sdk/core/serverless_lambda_context.py
chrisradek: Could that just be documented though, that calling put_segment in a Lambda context ends up  creating subsegments? That's what the ServerlessLambdaContext is doing. Mostly I'm not sure why there needs to be a distinction between ServerlessLambda and Lambda, and I think that was what @haotianw465 was hinting at in an earlier comment.

I don't think we have to worry about the case wherein we would want a segment in Lambda to be represented differently than a local subsegment, but correct me if I'm wrong!

chanchiem: Hmmm in that case, the usage requirements would be different; What happens if a customer attempts to create a subsegment when no parent segment (mimic segment) has been created (even though the facade segment exists)? Do we automatically attach that to the facade segment or do we not permit it? What if the customer runs put_segment then end_subsegment? I don't know if there are right or wrong answers here, but I think it requires a larger discussion.

Wouldn't we want a behavior to be consistent throughout the SDKs? From my knowledge, I don't think that the other SDKs support begin_segment explicitly in Lambda. This would mean that in Python, you would be able to begin segments and end segments as expected, but the rules will be different for other SDKs.

chanchiem:
Revision 3 added
* Mimic Segments now perform no-op on segment-only methods to ensure that no breaking changes will be sent to the data plane.
* Added comments to docs to address this contract.
* Unit tests to ensure segment-only methods don't add data to the mimic segment. Serialization methods are also tested to ensure this.

chanchiem:
Revision 4 Added:
* Revised design; no more serverless context; simpler design that is changed at the higher level middleware calls.
* get/save_origin_trace_header moved from segment to entity so subsegments can also propagate the header information. This is automatically deleted upon transmission and is only used to get request contexts in middlewares before and after requests.
* Subsegment generated when the middleware intercepts incoming request
* Unit tests to ensure that these subsegments are added on top of the facade segment

aws_xray_sdk/core/serverless_lambda_context.py
chanchiem: Decision has been made to discuss this issue sometime in a future date. For right now, we'll keep it simple and have the middlewares only begin subsegments when they usually begin segments.

aws_xray_sdk/ext/django/middleware.py
```diff
@@ -46,12 +51,15 @@ def __call__(self, request):
...
         )
 
-        segment = xray_recorder.begin_segment(
-            name=name,
-            traceid=xray_header.root,
-            parent_id=xray_header.parent,
-            sampling=sampling_decision,
-        )
+        if self.in_lambda:
+            segment = xray_recorder.begin_subsegment(name)
```
chanchiem: Trace header information, including the sampling decision, is propagated through to the Facade Segment.

aws_xray_sdk/core/models/entity.py
```diff
@@ -228,6 +230,20 @@ def add_exception(self, exception, stack, remote=False):
         self.cause['exceptions'] = exceptions
         self.cause['working_directory'] = os.getcwd()
 
+    def save_origin_trace_header(self, trace_header):
+        """
+        Temporarily store additional data fields in trace header
+        to the segment for later propagation. The data will be
```
yogiraj07: to the "Entity" for later propagation ...

chanchiem: Done

Issue #131, opened 2019-02-15
SQLAlchemy patching does not correctly remove password component from URL


WFT:
This was noticed when using `flask_sqlalchemy`, but I believe it affects plain `SQLAlchemy` as well.

```python
app = Flask(__name__) 
app.config["SQLALCHEMY_DATABASE_URI"] = "postgresql://foo:bar@baz.us-east-1.rds.amazonaws.com:2834/thing"
db = XRayFlaskSqlAlchemy(app)
```

Subsegments get recorded with the name `postgresql://foo:***@baz.us-east-1.rds.amazonaws.com:2834/thing`. Note that the password text is not present in the subsegment name (woohoo!), but it is composed of invalid characters, resulting in an error log message:
```
Removing Segment/Subsugment Name invalid characters from postgresql://foo:***@baz.us-east-1.rds.amazonaws.com:2834/thing
```

I believe that the issue lies here: https://github.com/aws/aws-xray-sdk-python/blob/dcb08012532d12403df0ec698c7158ab98a45878/aws_xray_sdk/ext/sqlalchemy/util/decorators.py#L104-L107

The URL is parsed and the password stripped, but the old URL is assigned to `safe_url`. The method `ParseResult._replace` returns a new `ParseResult` but does not modify the initial `ParseResult`.

Because the fix seems to be so simple, I'll put up a PR before hearing back, but please let me know if there's something I've misunderstood about this.

Configuration:
aws-xray-sdk-python == 2.3.0
Python 3.6

chanchiem:
Thank you very much for your contribution! I looked at your PR and it looks good to me. I have approved it. Very nice find :)

PR #132, opened 2019-02-15
Correctly remove password component from SQLAlchemy URLs, preventing …

WFT:
…invalid characters in SQLAlchemy subsegment names

*Issue #:* #131

*Description of changes:* 
The URL is parsed and the password stripped, but the old URL is assigned to `safe_url`. The method `ParseResult._replace` returns a new `ParseResult` but does not modify the initial `ParseResult`. This fixes the problem I described in the issue linked above where SQLAlchemy will replace the password component of SQLAlchemy URL with `***` which is then used as a Xray subsegment name, despite being invalid characters.

This is not a new feature, but a small fix to existing code which seems intended to do this in the first place. It seems the sort of thing which should require a test, but I'm not really sure how to add one that tests this since all the tests I can see use `sqlite:///:memory:` which as far as I know cannot have a password.

When I run the tests locally before this patch I have 5 failures in `tests/test_patcher.py`, when running after this patch, I get the same 5 failures.

### DIFF
--- a/aws_xray_sdk/ext/sqlalchemy/util/decorators.py
+++ b/aws_xray_sdk/ext/sqlalchemy/util/decorators.py
@@ -104,7 +104,7 @@ def parse_bind(bind):
             # Strip password from URL
             host_info = u.netloc.rpartition('@')[-1]
             parts = u._replace(netloc='{}@{}'.format(u.username, host_info))
-            safe_url = u.geturl()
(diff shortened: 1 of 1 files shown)


chanchiem:
Thank you very much for your contribution! This looks good to me. Very nice find :)

WFT:
Thank you!

Issue #134, opened 2019-02-21
Poller threads block main thread from exiting


bendog:
My script appeared to be completing all of it's steps, but wouldn't exit.
I set logging to debug and it appears that aws-xray-sdk is preventing the script from exiting.
It continuously logs this line forever.
`target_poller.py->_do_work:36 - DEBUG - There is no sampling rule statistics to report. Skipping`

I'm assuming there is something wrong with the config of my system as the aws xray console doesn't appear to have any traces showing up, however this still feels like a bug to me.

using: `aws-xray-sdk==2.3.0`
on EC2 (ubuntu 18.04) with `AWS X-Ray daemon version: 3.0.0`

bendog:
adding sampling=False resolved this.

```xray_recorder.configure(
    # daemon_address='127.0.0.1:2000',
    sampling=False,
    service=f'sms_eventhandler-{settings.ENVIRONMENT}',
    plugins=('EC2Plugin',)
)```

chanchiem:
Hey it looks like this issue is happening because the poller is having trouble obtaining your centralized sampling rules set in the X-Ray console. When you disabled sampling, the target poller was never actually started. The design of the target poller is to use a separate thread to do its work. The reason why it was stalling is because this separate thread was active in a spin loop to poll the service for sampling rules. In the case of disabling, the thread was never created so exiting was done gracefully

The reason it would log the issue you were having is because it was unable to obtain the sampling rules from the service. Do you have it configured through your console? Please refer here on how to configure it: https://docs.aws.amazon.com/xray/latest/devguide/xray-console-sampling.html

bendog:
I've only got the default catch-all sampling rule configured, I had assumed this should catch it.
Perhaps the Daemon I have running on the local ec2 instance can't access those rules, I might need to look into my IAM role access.

bendog:
I've updated my install script for xray daemon to sepecifically add the `Region: "ap-southeast-1"` entry to `/etc/amazon/xray/cfg.yaml`.
This seems to be at least get my events showing up in the xray console, I did previously have `AWS_DEFAULT_REGION=ap-southeast-1` in `/etc/environment` however it looks like this wasn't part of the xray service environment... 

```bash
# XRAY
curl -o /home/ubuntu/aws-xray-daemon-3.x.deb  https://s3.dualstack.us-east-2.amazonaws.com/aws-xray-assets.us-east-2/xray-daemon/aws-xray-daemon-3.x.deb
[...]

PR #144, opened 2019-02-28
Poller threads block main thread from exiting bug #134

chanchiem:
*Issue #, if available:*
#134 

*Description of changes:*
Set poller threads as daemon threads so they dont block the main thread on exit. 

### DIFF
--- a/aws_xray_sdk/core/sampling/rule_poller.py
+++ b/aws_xray_sdk/core/sampling/rule_poller.py
@@ -19,7 +19,9 @@ class RulePoller(object):
         self._connector = connector
 
     def start(self):
-        threading.Thread(target=self._worker).start()
+        poller_thread = threading.Thread(target=self._worker)
+        poller_thread.daemon = True
+        poller_thread.start()
 
     def _worker(self):
         frequency = 1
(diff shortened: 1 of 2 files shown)


chanchiem:
Minimal example to reproduce bug:
```

xray_recorder.configure(sampling=True)
xray_recorder.begin_segment("SomeSegment")
xray_recorder.end_segment()

exit() # Blocks forever
```

aws_xray_sdk/core/sampling/rule_poller.py
```diff
@@ -19,7 +19,9 @@ def __init__(self, cache, connector):
         self._connector = connector
 
     def start(self):
-        threading.Thread(target=self._worker).start()
+        poller_thread = threading.Thread(target=self._worker)
+        poller_thread.daemon = True
```
yogiraj07: Testing to consider:
1) Checking whether pollers start normally on starting application
2) If any exception occurs on any of the pollers, it doesn't break the application and the exception is caught. The poller doesn't die and continues to fetch information on next iteration.
3) The issue this solves the problem, i.e on terminating the application, the poller thread now terminates.

Please let me know if we have expected results.

chanchiem: 1. Pollers only start if sampling=True and beginSegment/beginSubsegment is called. This has always been the behavior, and the change hasn't affected it from my own tests.
2. Yes, on an exception, it doesn't break the application. In the _worker() method of the pollers, when any exception is caught, it simply logs it and retries the same cycle.
3. I had tested that this fix solves the problem using the minimal configuration shown in the comment I had posted. Prior to the fix, main thread is blocked. After the fix, main thread is no longer blocked and so the poller threads actually terminate.

Issue #155, opened 2019-05-17
SegmentNotFoundException was raised when AWS_XRAY_SDK_ENABLED=false and use/call any method of xray_recorder


luzfcb:
Hello, what is the correct way to disable X-Ray SDK without having to remove from my code the calls to the xray_recorder methods?

I'm using in a Django project that runs on Lambda, however, for local development or in Codebuild, I'd like to deactivate it, but if I disable the SDK by configuring the `AWS_XRAY_SDK_ENABLED` environment variable to `false` I get this error:

```
aws_xray_sdk.core.exceptions.exceptions.SegmentNotFoundException: cannot find the current segment/subsegment, please make sure you have a segment open
```

My project use 
```
boto3==1.9.144
botocore==1.12.145
aws-xray-sdk==2.4.2
django==2.1.8
```

chanchiem:
Thank you for reporting the bug! The intended behavior is to have it not throw exceptions through any calls to begin/end_segment or begin/end_subsegment no matter the order that is called.

It looks like we added this behavior in the `record_subsegment()` method here https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/recorder.py#L410 but not in the begin_subsegment method.

The solution to this would be to have the check written in `record_subsegment()` done at the `begin_subsegment()` method. We also need explicit checking for the end_segments and end_subsegment methods to ensure no exceptions are thrown.

luzfcb:
@chanchiem AWS have any deadlines for solving problems like this? Did you have a chance to look at my pr ()? Since `aws cloudformation package` does not work properly if there are packages installed directly from github, I can not use my fork in Lambda, and this becomes a blocker for me.

chanchiem:
Hey, I haven't had a chance to look at it yet. I'll prioritize this PR and will get it reviewed soon. Thank you very much for your contribution :)

chanchiem:
Took a look. Made a few comments about it. It's looking great so far!

chanchiem:
Customer Experience when SDK is disabled:
The following recorder APIs should not throw exceptions regardless of the order in which the APIs are called.
* `begin_subsegment()`
* `end_subsegment()`
* `end_segment()`
* `current_segment()`
* `current_subsegment()`

chensjlv:
+1 for a fix

we've already forked and patched this error internally with limited understanding of how xray works internally, but we'd like to see it gets fixed officially on the AWS side.

[...]

PR #158, opened 2019-05-20
Create DummySegment if SDK is disabled

luzfcb:
*Issue #, if available:*
#155

*Description of changes:*

To avoid get an `SegmentNotFoundException`, if SDK is disabled create a DummySegment before run `begin_subsegment`, `end_subsegment` and `end_segment`


luzfcb:
The tests fail but are unrelated to my changes

chanchiem:
* I'd recommend breaking down these boilerplate lines into a helper method for code clarity.
* Since we have this check in the begin/end subsegment methods now, we probably don't need the check anymore in record_subsegment().

chanchiem:
I'd also like to see unit tests added to make sure these methods behave as expected when running disabled.

```
xray_recorder.begin_subsegment() # Should not throw an error
```

``
xray_recorder.begin_segment()
xray_recorder.end_segment()
xray_recorder.begin_subsegment() # Should not throw an error
```
These are just a few test cases that I could think of.

**Comment by chanchiem** [comment]
Should we also force create a dummy entity for `xray_recorder.current_segment()` and `xray_recorder.current_subsegment()` if none exist? What would make most sense here? 

On the one hand, If the SDK is disabled, we wouldn't want customers to catch exceptions or even worry about any trickling errors that could possibly arise in their middlewares. For example if a middleware expects an entity to exist but it doesn't, but it calls `current_segment()` anyways.

On the other hand, they are expecting nothing to throw exceptions, although it would be a valid exception since they never created any entities, and this could indicate a bug in their application code.

I'm leaning towards the former: to return (and not store) a dummy entity if none exist.

**Comment by luzfcb** [comment]
> Should we also force create a dummy entity for xray_recorder.current_segment() and xray_recorder.current_subsegment() if none exist? What would make most sense here?

What's the right way to do this? if SDK is disabled, [`get_trace_entity `](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/recorder.py#L304) will log an error and will confuse the user

With my code, I still receive an error messages when I run Django manager.py

> I'm leaning towards the former: to return (and not store) a dummy entity if none exist.

I think that if SDK is disabled then the code executed in aws-xray-sdk should be as minimal as possible to avoid unnecessary processing.

**Comment by luzfcb** [comment]
Due to my knowledge of how these library works is infinitely less than your, please be free to ignore my pull-request, and make whatever corrections you feel necessary. I'll be happy to test your pull-request on my code on my staging server.

**Comment by luzfcb** [comment]
Question: Correct me if I'm wrong, but [`AWS_XRAY_SDK_ENABLED`](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/sdk_config.py#L30) seems to be evaluated when the python process the [`__init__.py`](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/__init__.py).  I think that the evaluation of this should be delayed for the first call of is `sdk_enabled`, That makes any sense?

**Comment by chanchiem** [comment]
> What's the right way to do this? if SDK is disabled, [`get_trace_entity `](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/recorder.py#L304) will log an error and will confuse the user

What I'm thinking is perhaps instead of even calling the get_trace_entity() method, it checks to see if the SDK is disabled, and if it is, simply do this: `return DummySegment(DEFAULT_DISABLED_NAME)` where `DEFAULT_DISABLED_NAME` can be a default name to signify a disabled entity. 

This is a suggestion to cover the case when a customer had disabled the SDK (and therefore would not want any exceptions to occur within their application). Since a customer's code can have any variation of statements, we want to make sure that we add something that's resilient to most use cases. For example, if a customer had disabled their SDK, and ran the following without first creating an entity: 
```
segment = xray_recorder.current_segment()
segment.put_annotation("test", "value")
```

I think this should still atleast execute and not throw a SegmentMissingException.

> I think that if the SDK is disabled then the code executed in aws-xray-sdk should be as minimal as possible to avoid unnecessary processing.

Agreed. We definitely want to minimize the execution paths of X-Ray when it's disabled.


> @chanchiem Due to my knowledge of how these library works is infinitely less than your, please be free to ignore my pull-request, and make whatever corrections you feel necessary. I'll be happy to test your pull-request on my code on my staging server.

I think your pull-request brings in a highly valuable fix to the feature that we weren't able to think about when writing it. I'll be happy to work with you some more to get this out, especially seeing how you've been blocked on this. I'll make it my priority.

> Question: Correct me if I'm wrong, but [`AWS_XRAY_SDK_ENABLED`](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/sdk_config.py#L30) seems to be evaluated when the python process the [`__init__.py`](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/__init__.py). I think that the evaluation of this should be delayed for the first call of is `sdk_enabled`, That makes any sense?

I think it can work both ways, but what's the main advantage to having it set upon the first call? 

The use case we were thinking about was that the environment variable should be set before the application is started; one of the top uses for this feature was to allow customers to run unit tests without having X-Ray throw exceptions because of missing segments that are usually generated automatically by the middlewares. The idea was to have them set the environment variable to disable the SDK and then run the test.

**Comment by chensjlv** [comment]
+1 for this fix

**Comment by chanchiem** [comment]
> +1 for this fix
> 
> we've already forked and patched this error internally, but we'd like to see it gets fixed officially on the AWS side.

The bug that this pull request attempts to solves indicates a deeper issue that we are currently investigating so that we can push out a fix that would encapsulate these use cases. Please stay tuned :)

File: aws_xray_sdk/sdk_config.py [hunk 1]
```diff
@@ -27,13 +27,17 @@ class SDKConfig(object):
...
+
+    __SDK_ENABLED = None
 
     @classmethod
     def sdk_enabled(cls):
         """
         Returns whether the SDK is enabled or not.
         """
+        if cls.__SDK_ENABLED is None:
+            cls.__SDK_ENABLED = str(os.getenv(cls.XRAY_ENABLED_KEY, 'true')).lower() != 'false'
```
luzfcb: why not something like
```
from distutils.util import strtobool
#...
            cls.__SDK_ENABLED = strtobool(os.getenv(cls.XRAY_ENABLED_KEY, 'true'))
```

`strtobool` source code:

https://github.com/python/cpython/blob/daf62627518ad97ce66a48c49496aa0573cf0731/Lib/distutils/util.py#L306-L319

chanchiem: Great suggestion! I'll add that to the next revision.

chanchiem:
The original design in begin_segment() caused the recorder to always generate dummy segments and store it in the context. It also didn't allow customers to call `begin_subsegment()` without first calling `begin_segment()`. This can be unexpected behavior for example if customers rely on instrumented frameworks to generate the segment.

As a result, we chose to make it so that all operations at the recorder level are either a no-op or returns a dummy entity without manipulating any components such as the sampler or context.

The tradeoff here is that data retrieval from the context would not work. This was a decision made because of the potential for memory leaks. If a customer had called `begin_subsegment()`, it would implicitly call `begin_segment()` and store a dummy segment into the context. If the customer's application code continually calls`begins_subsegment()`/`end_subsegment()` pairs but never calls `end_segment()` there would be a monotonically increasing subsegments list for that parent segment.

aws_xray_sdk/sdk_config.py
luzfcb: okay, `strtobool` lie because it returns 1 and 0, not True or False, and this can lead to errors because `1 is True` is false.

Maybe
```
            cls.__SDK_ENABLED = bool(strtobool(os.getenv(cls.XRAY_ENABLED_KEY, 'true')))
```

chanchiem: Done

chanchiem:
Hey, I've been pulled away from working on this for the time being and would be able to give some guidance in case anyone wants to take over the PR.

srprash:
Hi @luzfcb 
Thanks for the PR. I am currently looking into it. I think the approach we can take with disabling the SDK and still have it working as if nothing has changed could be to create a DummySegment and DummySubsegment and preserve them in the context as we do with normal Segment and Subsegment. 
Let me know how you feel about it. And if its okay with you, I would like to continue the work on this branch so that we have historical context of the fix.
Thanks.

Issue #159, opened 2019-06-03
Investigate why test_psycopg2 is failing with "UndefinedColumn" instead of "ProgrammingError"


chanchiem:
The test for some reason now is failing. This is aside from any recent changes that we've made.  

```
...
        exception = subsegment.cause['exceptions'][0]
>       assert exception.type == 'ProgrammingError'
E       AssertionError: assert 'UndefinedColumn' == 'ProgrammingError'
E         - UndefinedColumn
E         + ProgrammingError
tests/ext/psycopg2/test_psycopg2.py:147: AssertionError
...
```

We re-ran the test for a build that had previously worked and now it appears to be failing. This issue is  made to investigate this and will be updated when further information is found.

Please refer to this [link](https://github.com/aws/aws-xray-sdk-python/pull/157#issuecomment-497534572) for the build re-run of a previously passing build.

chanchiem:
In version 2.8.0 of Psycopg2, they announced the following:
`Added errors module. Every PostgreSQL error is converted into a specific exception class (ticket #682).`

This means that they added granular error exception classes, so the exception that we originally tested for (`ProgrammingError`) is now a base class exception for what was previously tested. In particular, the test that we originally wrote tested for a bad query whose database column doesn't exist, hence why we were seeing UndefinedColumn exceptions.

chanchiem:
Merged

PR #163, opened 2019-06-11
Fixes unit test for newer versions of psycopg2.

chanchiem:
Version 2.8.0 of Psycopg2 introduces granular errors for queries. This commit fixes a unit test to assert on the specific error instead of the now-updated base class. Functionally, the patcher still works fine with the latest versions of Psycopg2.

The previous matching class was ProgrammingError. For the specific test, it is now UndefinedColumn, which is a subclass of ProgrammingError.

*Issue #, if available:*
#159 

*Description of changes:*
Modified test to check for UndefinedColumn instead of ProgrammingError.

### DIFF
--- a/tests/ext/psycopg2/test_psycopg2.py
+++ b/tests/ext/psycopg2/test_psycopg2.py
@@ -144,7 +144,7 @@ def test_execute_bad_query():
     assert sql['database_version']
 
     exception = subsegment.cause['exceptions'][0]
-    assert exception.type == 'ProgrammingError'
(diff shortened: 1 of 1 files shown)


yogiraj07:
Sanity check : Do we need to do modification in AWS SDK handler w.r.t to this change?

chanchiem:
Nope this was a pure test error. Basically, Psycopg2 was updated so that they have more exceptions, so the only change that this'll do is that there are more types of exceptions that can be caught and stored in the segment.

PR #178, opened 2019-10-09
Faster Implementation of Wildcard Matching and added test file for wildcard matching

bhautikpip:
*Issue #, if available:*

*Description of changes:*

1. Faster Implementation of Wildcard matching 
2. Added test file for wildcard_match function

I have tested this change with the newly added test function.

### DIFF
--- a/aws_xray_sdk/core/utils/search_pattern.py
+++ b/aws_xray_sdk/core/utils/search_pattern.py
@@ -20,7 +20,7 @@ def wildcard_match(pattern, text, case_insensitive=True):
         return True
 
     # If elif logic Checking different conditions like match between the first i chars in text
-    # and the first j chars in pattern, checking pattern has '?' or '*' also check for case_insensitivity
+    # and the first p chars in pattern, checking pattern has '?' or '*' also check for case_insensitivity
     # iStar is introduced to store length of the text and i, p and pStar for indexing
     i = 0
     p = 0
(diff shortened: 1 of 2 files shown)


chanchiem:
It looks like code coverage decreased. Can we make sure all code paths are executed in your tests? Additionally, it would be nice if we could get a high level description of the algorithm in the comments for readability purposes.

And can we also confirm that these tests still pass before this change and after the change?

bhautikpip:
Tests have passed in both the version before this change and after this change. Moreover, I have added description of the algorithm. Initially code coverage was decreased because in the older version we are using simple_wildcard_change which we do not require in this version so I have removed that function.

aws_xray_sdk/core/utils/search_pattern.py
```diff
@@ -12,74 +12,47 @@ def wildcard_match(pattern, text, case_insensitive=True):
...
 
     # Check the special case of a single * pattern, as it's common
     if pattern == '*':
         return True
 
-    if case_insensitive:
-        pattern = pattern.lower()
-        text = text.lower()
+    # If elif logic Checking different conditions like match between the first i chars in text
+    # and the first j chars in pattern, checking pattern has '?' or '*' also check for case_insensitivity
```
chanchiem: there's no j; it should be a p

bhautikpip: Thanks for the correction! Committed that change.

PR #182, opened 2019-11-13
Add an option to force a normal context on Lambda

fjfnaranjo-wavelt:
Detect in `check_in_lambda` if an environment variable (SDK config level) is set and return early if its value is `normal`.

This will make the recorder use a plain `Context` object, instead of a `LambdaContext` object.

The rationale for this: I need to report the URL in the top segment for a server-less application I'm working on. I want to analyse some GraphQL queries in the console and I need their name in the URL, to classify them. If I enable X-Ray reporting in the Lambda console, a top segment is created and I'm unable to change its URL field (as said in the docs). But if I disable X-Ray to create the segment by myself (loosing cold-start info is ok in my case) the SDK still thinks a LambdaContext should be used. With a `LambdaContext` you cannot create top segments. This change allows the user of the SDK to override the kind of context used.

### DIFF
--- a/aws_xray_sdk/core/lambda_launcher.py
+++ b/aws_xray_sdk/core/lambda_launcher.py
@@ -21,6 +21,11 @@ def check_in_lambda():
     Return None if SDK is not loaded in AWS Lambda worker.
     Otherwise drop a touch file and return a lambda context.
     """
+
+    override_context = os.getenv('XRAY_OVERRIDE_CONTEXT')
+    if override_context == 'normal':
+        return None
+
     if not os.getenv(LAMBDA_TASK_ROOT_KEY):
         return None
 


chanchiem:
Hey,

while this work around may work for the time being, we do not recommend using this for the long run as any changes made to the lambda service may break this implementation. Therefore, we will need to decline this pull request. We still really appreciate your contributions and would wish to possibly revisit this in the future if changes are made in the long run to allow this support.

We currently do support the Serverless Architecture, where if your API is fronted with API Gateway and computed using Lambda, the downstream calls made from your lambda function will all be connected.

But it seems like you're attempting to aggregate downstream GraphQL queries. Have you considered using filter expressions to help aggregate your queries? If you need more help with this, we recommend opening a github issue for further discussions.

fjfnaranjo-wavelt:
Thanks for the feedback!

What we wanted with this is to have the option to change the URL metadata field. This way we can choose 'URL' in the 'Group by' selector under the 'Trace overview' header in the 'Traces' section.

We can't find a way to use filters to have this information (percentage of traces + average response time). And we can't use the default X-Ray integration in Lambda because we can not edit the top level segment (the one used to populate the URL field in the X-Ray interface).

The problem with `check_in_lambda` is that it uses an environment variable available in all Lambda functions. Even if we wanted to create our own segment, we can not do it because the SDK always _think_ that a segment was created by the Lambda integration and disallow the creation of other top segments (see `LambdaContext`).

We will stick to our workaround because we need this information now. We are studying other solutions more focused on APMs but we want to have this option available in X-Ray.

Issue #191, opened 2019-12-05
Downstream http requests should use the hostname as a subsegment name


chanchiem:
Requests, HttpLib, and Aiohttp should all generate subsegments whose name is assigned as the hostname, rather than the full URL path.

Reproduction:
```
patch_all()
xray_recorder.begin_segment("segment")
requests.get("https://aws.amazon.com/some/url/endpoint")
xray_recorder.end_segment()
```

observe that the subsegment name is `https://aws.amazon.com/some/url/endpoint` instead of "aws.amazon.com`

PR #192, opened 2019-12-05
Downstream Http Calls should use hostname rather than full URL as subsegment name

chanchiem:
*Issue #, if available:*
#191 

*Description of changes:*
Subsegment names generated by requests, httplib, aiohttp should  be the hostname, not the full URL

Added unit tests for requests, httplib, aiohttp to validate this.

### DIFF
--- a/aws_xray_sdk/ext/util.py
+++ b/aws_xray_sdk/ext/util.py
@@ -14,6 +14,7 @@ else:  # Python 2 and below
 
 first_cap_re = re.compile('(.)([A-Z][a-z]+)')
 all_cap_re = re.compile('([a-z0-9])([A-Z])')
+UNKNOWN_HOSTNAME = "UNKNOWN HOST"
 
 
 def inject_trace_header(headers, entity):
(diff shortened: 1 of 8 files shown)


aws_xray_sdk/ext/util.py
```diff
@@ -118,6 +124,14 @@ def strip_url(url):
     return url.partition('?')[0] if url else url
 
 
+def get_hostname(url):
+    if url is None:
+        return None
+    url_parse = urlparse(url)
+    hostname = url_parse.hostname
+    return hostname if hostname else url  # If hostname is none, we return the regular URL; indication of malformed url
```
c1tadel: It this fall-through behavior consistent with other SDKs? If we ask an HTTP client for a URL and get a malformed answer I'm not sure why we'd want to pass it back to the trace -- seems like a repeat of the issue this is intended to solve.

awssandra: Node does "unknown host" - https://github.com/aws/aws-xray-sdk-node/blob/master/packages/core/lib/patchers/http_p.js#L67

chanchiem: this is in the case where the URL passed in might not necessarily be a parseable URL. Since customers can randomly invoke something like: `requests.get("INVALID_DOMAIN_NAME")`, we'd still want to be able to capture these types of errors and indicate failure in the subsegment. Otherwise, the subsegment won't be generated at all.

The behavior on our SDKs are mixed; some of them throw NP exceptions because the hostname technically doesn't exist, and therefore doesn't produce anything (which seems like a bug to me).

With that being said, I think doing something similar to Node would be the best of both worlds. Using a known "bad hostname" name, and then still capturing the exception in the subsegment.

PR #95, opened 2018-09-12
Fix psycopg2 register type

jaysonsantos:
This patch is intended to fix problems when you try to register extensions like uuid on psycopg2.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #198, opened 2020-01-10
Unwrap context provided to psycopg2.extensions.quote_ident

labichn:
This PR fixes a bug in the use of `psycopg2.extensions.quote_ident` when psycopg2 is patched by this SDK. The bug is similar to the one fixed by #95, where a wrapped cursor or connection [causes a type error](https://github.com/psycopg/psycopg2/blob/master/psycopg/psycopgmodule.c#L180).

The [new test case](https://github.com/UnitedIncome/aws-xray-sdk-python/blob/a83b482530b72bf49235568d2dfc252422de3f32/tests/ext/psycopg2/test_psycopg2.py#L164) gives a minimal example of the bug, which fails against the current master ([4689220](https://github.com/aws/aws-xray-sdk-python/tree/46892208561107405d68e64b3423eea54e7b8127)).

### DIFF
--- a/aws_xray_sdk/ext/psycopg2/patch.py
+++ b/aws_xray_sdk/ext/psycopg2/patch.py
@@ -18,6 +18,11 @@ def patch():
         'register_type',
         _xray_register_type_fix
     )
+    wrapt.wrap_function_wrapper(
+        'psycopg2.extensions',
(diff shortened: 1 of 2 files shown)


chanchiem:
Thank you very much for your contribution :)

Issue #201, opened 2020-01-29
"failed to send data to X-Ray daemon." -> "Message too long"


tobiasmcnulty:
We've enabled the `record_full_document` parameter for the pymongo patch method (https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/pymongo/patch.py#L12-L14) and are seeing the below exception when attempting to record a large query/response from mongo.

Would it be possible to update the UDP emitter to truncate the data or split it up so we can see at least some of it in the console?

Many thanks!

```
failed to send data to X-Ray daemon.
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/aws_xray_sdk/core/emitters/udp_emitter.py", line 63, in _send_data
    self._port))
OSError: [Errno 90] Message too long
```

tobiasmcnulty:
I should add, we're also seeing events with no segments in the console, which is maybe related?

![Screen Shot 2020-01-29 at 12 51 30 PM](https://user-images.githubusercontent.com/160132/73382783-2adee600-4296-11ea-833a-c83c95a4b6cd.png)

(ID has been obfuscated.)

awssandra:
Hi tobiasmcnulty,

I'd suggest using the streaming threshold in the recorder first. The idea is that subsegments can be batched and sent independently of their segment, as long as they're complete. This can help alleviate the UDP size limit when working with large segment documents.

[Relevant Note](https://github.com/aws/aws-xray-sdk-python/blob/d081441dd268532c03813adf21ab85c2c2749659/aws_xray_sdk/core/recorder.py#L127)
[More Notes](https://github.com/aws/aws-xray-sdk-python/blob/d081441dd268532c03813adf21ab85c2c2749659/aws_xray_sdk/core/recorder.py#L130)

We have similar logic on Lambda, where we [force every subsegment to send when complete](https://github.com/aws/aws-xray-sdk-python/blob/d081441dd268532c03813adf21ab85c2c2749659/aws_xray_sdk/core/recorder.py#L60)

So, I'd expect something like the following to work...
`
from aws_xray_sdk.core import xray_recorder

xray_recorder.configure(
    //... configs here
    streaming_threshold=0
    //Magic number here, default is 30. Setting to 0 will force a subsegment to send as soon as it's closed.
)
`

dpoirier:
[...]

PR #205, opened 2020-02-25
streaming_threshold not None check. Also dropping `loop` parameter from aiobotocore tests.

srprash:
*Issue #, if available:* [201](https://github.com/aws/aws-xray-sdk-python/issues/201)

*Description of changes:*
While setting the streaming_threshold in `xray_recorder.configure()`, if the value is 0, the condition evaluates to false if not checked for None.


tox.ini
```diff
@@ -29,9 +29,13 @@ deps =
     py{27}: enum34
 
     # Python3.5+ only deps
-    py{35,36,37,38}: aiohttp >= 3.0.0
+    py{35,36,37,38}: aiohttp >= 3.3.1
```
bhautikpip: Have they deprecated support for aiohttp 3.0.0 ?

srprash: I don't think so. But the aiobotocore setup file requires aiohttp >= 3.3.1 so I made that change too. I'm not sure if it has any implications.

tox.ini
```diff
@@ -29,9 +29,13 @@ deps =
...
     py{35,36,37,38}: pytest-aiohttp
-    py{35,36,37,38}: aiobotocore >= 0.10.0
+    py{35,36,37,38}: botocore >= 1.15.3,<1.15.16
+    # aiobotocore 0.12.0 is not available for python 3.5
+    py{35}: aiobotocore >= 0.10.0
```
bhautikpip: I think it will look for latest version so you can add this case with other cases. py{35,36,37,38} >= 0.12.0

tox.ini
```diff
@@ -29,9 +29,13 @@ deps =
     py{27}: enum34
 
     # Python3.5+ only deps
-    py{35,36,37,38}: aiohttp >= 3.0.0
+    py{35,36,37,38}: aiohttp >= 3.3.1
     py{35,36,37,38}: pytest-aiohttp
-    py{35,36,37,38}: aiobotocore >= 0.10.0
+    py{35,36,37,38}: botocore >= 1.15.3,<1.15.16
```
bhautikpip: <1.15.16 means we will have to keep updating this version number in the case of future updates with botocore. I think botocore >= 1.15.3 should be good.

tox.ini
srprash: actually we may not even need this case at all. having py{35,36,37,38} >= 0.10.0 should pull in whatever latest version is supported for each runtime.

tox.ini
srprash: You're right. Probably don't need this one as well.

tox.ini
bhautikpip: Yeah. That's what I am suggesting. we can keep 0.10.0

tox.ini
bhautikpip: Okay!

tox.ini
```diff
@@ -98,4 +98,4 @@ commands =
 
 [flake8]
 max-line-length=120
-exclude=tests
+exclude=tests
```
bhautikpip: Can you add a new line here?

Issue #168, opened 2019-07-28
Subsegment name serialization causing MissingName error


Jacob-Doetsch:
XRay Daemon reports the following error:
```bash
xray-daemon_1  | 2019-07-28T17:35:30Z [Info] Successfully sent batch of 1 segments (0.481 seconds)
xray-daemon_1  | 2019-07-28T17:35:30Z [Error] Unprocessed segment: {
xray-daemon_1  |   ErrorCode: "MissingName",
xray-daemon_1  |   Id: "3bdcccc4799c57fa",
xray-daemon_1  |   Message: "Invalid subsegment. ErrorCode: MissingName, Cause: null"
xray-daemon_1  | }
xray-daemon_1  | 2019-07-28T17:35:30Z [Warn] Batch that contains unprocessed segments
```

Simple setup. Django app, using middleware. setup includes:
```python
XRAY_RECORDER = {
    'AWS_XRAY_DAEMON_ADDRESS': 'xray-daemon:2000',
    'AWS_XRAY_TRACING_NAME': 'TheNextBigThing',
    'AWS_XRAY_CONTEXT_MISSING': 'LOG_ERROR'
}

INSTALLED_APPS = [
    ...
    'aws_xray_sdk.ext.django',
]

MIDDLEWARE = [
    'aws_xray_sdk.ext.django.middleware.XRayMiddleware',
    ...
]
```

I think subsegments' subsegments' names are not being serialized correctly. Below is a snippet from the XRay Daemon payload.Notice the `"name": {"py/newargs": ["core/snippets/ga.html"]}` name of the subsegment, which should instead be a string according to the segment JSON schema.

```json
    ...
    "subsegments": [
        {
            ...
            "in_progress": false,
            "name": "account/login.html",
            "namespace": "local",
            "parent_id": "c572ac3f20a195dc",
            "subsegments": [
                {
                    "id": "69ef19799fec25ba",
                    "in_progress": false,
                    "name": {
                        "py/newargs": [
                            "core/snippets/ga.html"
                        ]
                    },
                    "namespace": "local",
                    "parent_id": "5ecd89473a50fa17",
                    "type": "subsegment"
                },
                ...
            ]
            ...
        }        
```

All SDK versions between 2.0.1 and 2.4.2 were tested and have the same issue. Same issue on Python 3.6 and 3.7 as well. Django version is 2.2.3.

chanchiem:
Hey,

[...]

PR #200, opened 2020-01-28
Add testing for current Django versions

ipmb:
refs #10 #85 

*Description of changes:*

Update tox so it will test against currently supported versions of Django.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #206, opened 2020-02-26
Add support for Django 2.0 to 3.0

kmichel-sereema:
This fixes a runtime issue with template names in subsegments (#168) and two failed imports during tests.

### DIFF
--- a/aws_xray_sdk/ext/django/templates.py
+++ b/aws_xray_sdk/ext/django/templates.py
@@ -1,6 +1,7 @@
 import logging
 
 from django.template import Template
+from django.utils.safestring import SafeString
 
 from aws_xray_sdk.core import xray_recorder
 
@@ -22,6 +23,10 @@ def patch_template():
         template_name = self.name or getattr(context, 'template_name', None)
         if template_name:
             name = str(template_name)
+            # SafeString are not properly serialized by jsonpickle,
+            # turn them back to str by adding a non-safe str.
+            if isinstance(name, SafeString):
+                name += ''
             subsegment = xray_recorder.current_subsegment()
             if subsegment:
                 subsegment.name = name
--- a/tox.ini
+++ b/tox.ini
@@ -21,9 +21,9 @@ deps =
     Flask-SQLAlchemy
     future
     # the sdk doesn't support earlier version of django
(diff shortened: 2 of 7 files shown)


kmichel-sereema:
Hi,

I fixed the issues I had when deploying xray with Django 3.0 but removed the commit in this PR enabling coverage with tox: I'm too confused about tox to create a correct configuration and only noticed #200 after pushing this code.

srprash:
Hi @kmichel-sereema 
Thanks for the PR. Your tests for Python 3.6+ are failing because of a change in aiobotocore 0.12.0 a couple of days back where the `loop` parameter was dropped from the `get_session` api. I have pushed the changes to our current aiobotocore tests in [this PR](https://github.com/aws/aws-xray-sdk-python/pull/205). Your tests should be passing once this PR is merged.

aws_xray_sdk/ext/django/templates.py
```diff
@@ -22,6 +23,10 @@ def xray_render(self, context):
         template_name = self.name or getattr(context, 'template_name', None)
         if template_name:
             name = str(template_name)
+            # SafeString are not properly serialized by jsonpickle,
+            # turn them back to str by adding a non-safe str.
+            if isinstance(name, SafeString):
```
srprash: Can you add a test case for this code path to maintain coverage?

kmichel-sereema: I added the test and updated the django version (the test wouldn't cover the case without updating django).

tox.ini
```diff
@@ -19,7 +21,9 @@ deps =
     Flask-SQLAlchemy
     future
     # the sdk doesn't support earlier version of django
-    django >= 1.10, <2.0
+    py{27,34}-!django2-!django3,django1: django >= 1.10, <2.0
```
srprash: You won't probably need to specify the supported django version for each runtime. If you just do
`py{27,34,35,36,37,38}: django >= 1.10`, tox should automatically pick up the highest version for each python runtime. Please correct me if I'm wrong.

kmichel-sereema: You're right, tox is smart enough to rely on pypi classifiers.

Thanks for the suggestion.

PR #217, opened 2020-04-11
asyncio.Task.current_task PendingDeprecation fix

### LINKED ISSUE #216: asyncio.Task.current_task PendingDeprecation warning
Getting a lot of `PendingDeprecation` warnings in the logs.

/usr/local/lib/python3.7/site-packages/aws_xray_sdk/core/async_context.py:64: PendingDeprecationWarning: Task.current_task() is deprecated, use asyncio.current_task() instead

Should be an easy fix.

Amertz08:
*Issue #, if available:*

Resolves #216 

*Description of changes:*

Change `asyncio.Task.current_task` invocation to `asyncio.current_task` to alleviate `PendingDeprecationWarning`


Amertz08:
Yeah so unfortunately simply changing the invocation won't work with 3.5 and 3.6. I think the 2 solutions would be to either do a `try/except` catch of an `AttributeError` and then fall back to the compatible version or use and if/else off of `sys.version_info`. The latter would be more clean imo.

bhautikpip:
Yes. I also think use of sys.version_info would be more cleaner.

Amertz08:
Looks like it is all back to green in the test suite. Not sure why the check isn't working.

aws_xray_sdk/core/async_context.py
```diff
@@ -1,7 +1,10 @@
 import asyncio
+import sys
 
 from .context import Context as _Context
 
+_GTE_PY37 = sys.version_info.minor >= 7
+
```
bhautikpip, line 7: I would also add condition which checks sys.version_info.major is 2 and 3 because in the case of major version 4.0 above condition would break.

Amertz08: Do we need a guard against 4? I think we do need to make sure its python3. I forgot python 2.7 had asyncio support. So I assumed no python2 code would call this code and if it did it would error for other reasons.

So basically just this.
```python
_GTE_PY37 = sys.version_info.major == 3 and  sys.version_info.minor >= 7
```

bhautikpip: Yep. I think this makes sense.

PR #226, opened 2020-05-20
IMDSv2 support for EC2 plugin

srprash:
*Issue #, if available:*

*Description of changes:*
For better security, EC2 provides a IMDSv2 endpoint to [fetch the instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html). This PR aims at fetching the metadata from v2 endpoint to begin with and fallback to v1 if the former fails.


aws_xray_sdk/core/plugins/ec2_plugin.py
```diff
@@ -17,16 +19,51 @@ def initialize():
...
+            return None
+        req = Request(url=url)
+        req.headers = headers
+        req.method = method
+        res = urlopen(req, timeout=1)
+        return res.read().decode('utf-8')
     except Exception:
-        runtime_context = None
-        log.warning("failed to get ec2 instance metadata.")
+        raise
```
willarmiros, line 69: If we're just raising these exceptions anyway does it need to be in try/except block?

aws_xray_sdk/core/plugins/ec2_plugin.py
```diff
@@ -17,16 +19,51 @@ def initialize():
...
+            runtime_context['availability_zone'] = do_request(url=IMDS_URL + "meta-data/placement/availability-zone-1")
+
+        except Exception as e:
+            runtime_context = None
+            log.warning("failed to get ec2 instance metadata from IMDSv1 due to {}".format(e))
+
+
+def do_request(url, headers=None, method="GET"):
+    if headers is None:
+        headers = {}
```
willarmiros, line 59: Can the default value of headers in the signature be `{}` instead of `None`?

aws_xray_sdk/core/plugins/ec2_plugin.py
```diff
@@ -17,16 +19,51 @@ def initialize():
...
+                                                    method="GET")
+
+        # get availability-zone metadata
+        runtime_context['availability_zone'] = do_request(url=IMDS_URL + "meta-data/placement/availability-zone",
+                                                          headers={"X-aws-ec2-metadata-token": token},
+                                                          method="GET")
+
+    except Exception as e:
+        # Falling back to IMDSv1 endpoint
+        log.warning("failed to get ec2 instance metadata from IMDSv2 due to {}. Falling back to IMDSv1".format(e))
```
willarmiros, line 43: Consider making this a debug statement since, to my understanding, not all instances have v2 enabled

aws_xray_sdk/core/plugins/ec2_plugin.py
```diff
@@ -17,16 +19,51 @@ def initialize():
...
+
+        try:
+            runtime_context = {}
 
-        r = urlopen('http://169.254.169.254/latest/meta-data/placement/availability-zone',
-                    timeout=1)
-        runtime_context['availability_zone'] = r.read().decode('utf-8')
+            runtime_context['instance_id'] = do_request(url=IMDS_URL + "meta-data/instance-id")
 
+            runtime_context['availability_zone'] = do_request(url=IMDS_URL + "meta-data/placement/availability-zone-1")
```
willarmiros, line 50: Just as a sanity check, even though `IMDS_URL` has latest in it, these requests will still go to v1 right?

tests/test_plugins.py
```diff
@@ -13,3 +14,33 @@ def test_runtime_context_available():
...
+
+@patch('aws_xray_sdk.core.plugins.ec2_plugin.do_request')
+def test_ec2_plugin(mock_do_request):
+    mock_do_request.side_effect = ['token', 'i-0a1d026d92d4709cd', 'us-west-2b',
+                                   Exception("Boom!"), 'i-0a1d026d92d4709ab',
+                                   'us-west-2a',
+                                   Exception("Boom v2!"), Exception("Boom v1!")]
+
+    ec2_plugin = get_plugin_modules(('ec2_plugin',))
+    for plugin in ec2_plugin:
```
willarmiros, line 27: Is this for loop necessary? It is somewhat confusing since it seems like you're only doing 1 iteration for the `ec2_plugin`

tests/test_plugins.py
```diff
@@ -13,3 +14,33 @@ def test_runtime_context_available():
     for plugin in plugins:
         plugin.initialize()
         assert hasattr(plugin, 'runtime_context')
+
+
+@patch('aws_xray_sdk.core.plugins.ec2_plugin.do_request')
+def test_ec2_plugin(mock_do_request):
```
willarmiros, line 20: Can we turn this into 3 unit tests, one for each condition it checks? That way in the future if one test breaks it will be easier to know what specifically is broken.

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: Yes, v2 is optional and v1 will always be there unless customer configures for v2 only. It's not up to the plugin to determine the configuration, therefore I feel that both v2 and v1 failure can be debug statement and not getting any metadata can be a warning statement.

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: You're right. Don't really need the try/except block

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: Yes. v2 is optional and v1 will always be present unless customer has configured v2 only. I think it would be better to have debugs for v2 or v1 failure and a warning for not being able to fetch the metadata at all.

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: Yes. Both v1 and v2 have same urls. Just that the v2 requires a header value.

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: As a python best practice, mutable values (like [] or {}) should not be used as argument defaults.

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: Correct. No need for the try/except block here.

tests/test_plugins.py
srprash: Good point

tests/test_plugins.py
srprash: Right. Don't really need the loop.

tests/test_plugins.py
srprash: Broke down into 3 tests

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: Added debugs and a warning

aws_xray_sdk/core/plugins/ec2_plugin.py
srprash: removed the try except block

PR #237, opened 2020-07-20
set pymysql version to 0.9.3 in case of 0.10 upgrade

wangzlei:
*Issue #, if available:*
PyMySQL upgrade from 0.9.3 to 0.10 on 18th July, the new version changes a lot includes some exception types changes which causes x-ray sdk unit test failed.
https://github.com/PyMySQL/PyMySQL/blob/master/CHANGELOG.md

*Description of changes:*
roll back PyMySQL version to 0.9.3.


srprash:
Would need to fix the unit test to test on lates version of pymysql

PR #239, opened 2020-08-04
Generalize error check for pymysql error type

srprash:
*Description of changes:*
In the pymysql v0.10.0, the error type is changed from InternalError to OperationalError. The test should be testing on the lates version of pymysql, therefore having a general check instead of checking specific error type would avoid test failure due to such breaking changes in the future.

### DIFF
--- a/tests/ext/pymysql/test_pymysql.py
+++ b/tests/ext/pymysql/test_pymysql.py
@@ -76,4 +76,4 @@ def test_execute_bad_query():
     assert sql['database_version']
 
     exception = subsegment.cause['exceptions'][0]
-    assert exception.type == 'InternalError'
+    assert exception.type is not None


michael-k:
You've beaten me by a couple of minutes. ;)

If anyone's interested, here's the PR that made the change in PyMySQL: PyMySQL/PyMySQL#823.  The `errno` is 1054 in this case, see https://dev.mysql.com/doc/mysql-errors/5.7/en/server-error-reference.html#error_er_bad_field_error

srprash:
> If anyone's interested, here's the PR that made the change in PyMySQL: [PyMySQL/PyMySQL#823](https://github.com/PyMySQL/PyMySQL/pull/823). The `errno` is 1054 in this case, see https://dev.mysql.com/doc/mysql-errors/5.7/en/server-error-reference.html#error_er_bad_field_error

Thanks for the deep dive and posting the actual errorno. :)

PR #252, opened 2020-11-26
Allow configuring different Sampler in Django App

lalvarezguillen:
This patch adapts `aws_xray_sdk.ext.django` to be able to configure `Sampler` type.

Because right now it uses `aws_xray_sdk.core.sampling.sampler.DefaultSampler`, and doesn't allow configuring an alternative. Which means that it's impossible to use local sampling rules with Django

### DIFF
--- a/tests/ext/django/test_settings.py
+++ b/tests/ext/django/test_settings.py
@@ -1,0 +1,16 @@
+from unittest import mock
+
+import django
+from aws_xray_sdk import global_sdk_config
+from django.test import TestCase, override_settings
+from django.conf import settings
+from django.apps import apps
+
+from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.core.sampling.sampler import LocalSampler
+
+
+class XRayConfigurationTestCase(TestCase):
+    def test_sampler_can_be_configured(self):
+        assert isinstance(settings.XRAY_RECORDER['SAMPLER'], LocalSampler)
+        assert isinstance(xray_recorder.sampler, LocalSampler)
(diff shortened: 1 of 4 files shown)


tests/ext/django/test_settings.py
```diff
@@ -0,0 +1,16 @@
...
+from django.conf import settings
+from django.apps import apps
+
+from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.core.sampling.sampler import LocalSampler
+
+
+class XRayConfigurationTestCase(TestCase):
+    def test_sampler_can_be_configured(self):
+        assert isinstance(settings.XRAY_RECORDER['SAMPLER'], LocalSampler)
```
lalvarezguillen, line 15: I would have loved to overwrite `settings.XRAY_RECORDER['SAMPLER']` here, to show the recorder working with different samplers.

But because `xray_recorder` is configured in the `ready` method of the app, and the `ready` method is only called once when the test suite is run, that's not easy to implement (cleanly) currently. (you can read more about it [here](https://stackoverflow.com/questions/31148172/django-override-setting-used-in-appconfig-ready-function), or verify with a `pdb.set_trace()` on the `ready` method)

srprash: I guess there is way to test different samplers by having different django settings and have it loaded using `settings.configure(default_settings=settings_with_LocalSampler)` and then calling `django.setup()` for each test case. More info [here](https://docs.djangoproject.com/en/3.1/topics/settings/). But I agree with you that it may not be clean. I think this is good for now since it tests what we want at this point.

srprash (approved):
Hi @lalvarezguillen 
Thanks for filling in this gap for the SDK. I'll update the [docs](https://docs.aws.amazon.com/xray-sdk-for-python/latest/reference/frameworks.html) to reflect the new settings parameter.

Issue #255, opened 2020-12-02
Deprecate support for python 2.7 and 3.4 in the next major version


srprash:
Python [2.7](https://www.python.org/doc/sunset-python-2/)(the last of python2) and [3.4](https://www.python.org/downloads/release/python-3410/#:~:text=10%20was%20released%20on%20March,new%20releases%20will%20be%20made.) have reached their end-of-life and are no longer supported.

The SDK should deprecate the support for these versions in the 3.0.0 release. FYI, the OpenTelemetry for Python [supports python 3.5 and above](https://github.com/open-telemetry/opentelemetry-python/blob/master/opentelemetry-api/setup.cfg#L37).

michael-k:
Python 3.5 has reached its EOL in September 2020: https://www.python.org/dev/peps/pep-0478/

OpenTelemetry for Python dropped support for Python 3.5 in https://github.com/open-telemetry/opentelemetry-python/commit/d21793d1249ee6e41180dc9a33755c6dcc0e94a1.

michael-k:
Python 3.6 has reached its EOL in December 2021: https://peps.python.org/pep-0494/

OpenTelemetry for Python is preparing to drop support for Python 3.6, see https://github.com/open-telemetry/opentelemetry-python/pull/2763.

michael-k:
@jj22ee has dropped support for Python < 3.7 in 2.12, see eg. eaffcc801138fcf44d540dbf500299285a453158.  But there's still work to do:
- The [classifiers](https://github.com/aws/aws-xray-sdk-python/blob/c57d293198251e276988440400140ac247c24606/setup.py#L31-L36) still suggest support for Python 2.7 and 3.4–3.6.
- [install_requires](https://github.com/aws/aws-xray-sdk-python/blob/c57d293198251e276988440400140ac247c24606/setup.py#L44-L49) includes dependencies for Python 2.7
- There's no python_requires to prevent pip from installing 2.12 on Python < 3.7.
- … (only had a rough look so far)

PR #260, opened 2020-12-11
Using GH actions for unit tests in place of travis

srprash:
Changing our CI to run unit tests through Github actions instead of Travis. Travis has been a little flaky in the past with jobs hanging to report back the status due this known [issue](https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154).

I played with a couple of options for this workflow and both have some caveats. 
First, I tried using the `tox -e py` command in the workflow to run tox tests as mentioned in the official GH [docs](https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python#running-tests-with-tox). But looks like it only works if you have simple tox environments like py27, py34, py 35 and so on, and not if you have environment like py27-default, py38-django22 which is what we have [here](https://github.com/aws/aws-xray-sdk-python/blob/master/tox.ini#L3-L7).
Then I tried this seemingly popular tox plugin [`tox-gh-actions`](https://pypi.org/project/tox-gh-actions/) which works with any tox environment. This works for us and I did some trial runs [here](https://github.com/srprash/aws-xray-sdk-python/runs/1535165085?check_suite_focus=true). The only problem with this plugin is that its not available for Python3.4 so we won't be able to run test for the `py34-default` environment.

I'm okay with holding out in merging this PR until our next major version where we [plan](https://github.com/aws/aws-xray-sdk-python/issues/255) to deprecate python2.7 and python3.4.


srprash:
Closing this PR in favor of a new one: https://github.com/aws/aws-xray-sdk-python/pull/322

PR #263, opened 2020-12-17
Add ability to ignore some requests from httplib

jonathangreen:
*Description of changes:*
This generalizes the test that the httplib patch was doing to ignore requests to xray being made by botocore. I am using cloudwatch logging in my application, and these calls occur outside of the normal application flow, so I don't have a segment open for them. 

This adds a new function `add_ignored` that allows requests to be ignored based on the `hostname`, `url` or `subclass`.  It defaults to ignoring the same things as before, but now additional requests can be ignored.

I also added some tests for the new functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -476,7 +476,7 @@ db = XRayFlaskSqlAlchemy(app)
 
 ### Ignoring httplib requests
 
-If you want to ignore certain httplib requests you can do so based on the hostname or URL that is being requsted. 
+If you want to ignore certain httplib requests you can do so based on the hostname or URL that is being requsted. The hostname is matched using the Python [fnmatch library](https://docs.python.org/3/library/fnmatch.html) which does Unix glob style matching.
 
 ```python
 from aws_xray_sdk.ext.httplib import add_ignored as xray_add_ignored
@@ -492,6 +492,15 @@ xray_add_ignored(urls=['/test-path', '/other-test-path'])
 
 # ignore requests to myapp.com for /test-url
 xray_add_ignored(hostname='myapp.com', urls=['/test-url'])
+```
+
+If you use a subclass of httplib to make your requests, you can also filter on the class name that initiates the request. This must use the complete package name to do the match.
+
+```python
+from aws_xray_sdk.ext.httplib import add_ignored as xray_add_ignored
+
+# ignore all requests made by botocore
+xray_add_ignored(subclass='botocore.awsrequest.AWSHTTPConnection')
 ```
 
 ## License
--- a/README.md
+++ b/README.md
@@ -473,6 +473,27 @@ XRayMiddleware(app, xray_recorder)
 db = XRayFlaskSqlAlchemy(app)
 
 ```
+
+### Ignoring httplib requests
+
+If you want to ignore certain httplib requests you can do so based on the hostname or URL that is being requsted. 
+
+```python
+from aws_xray_sdk.ext.httplib import add_ignored as xray_add_ignored
+
+# ignore requests to test.myapp.com
+xray_add_ignored(hostname='test.myapp.com')
+
+# ignore requests to a subdomain of myapp.com with a glob pattern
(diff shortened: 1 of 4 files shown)


srprash:
@jonathangreen Thanks for this contribution! I'll give it a quick try. Could you also add a section in the README with an example for using the `add_ignored` method for ignoring requests?

jonathangreen:
@srprash Thanks for reviewing! I've added some documentation to the README with an example of using the `add_ignored` method.

README.md
```diff
@@ -473,6 +473,27 @@ XRayMiddleware(app, xray_recorder)
 db = XRayFlaskSqlAlchemy(app)
 
 ```
+
+### Ignoring httplib requests
+
+If you want to ignore certain httplib requests you can do so based on the hostname or URL that is being requsted.
```
willarmiros, line 479: Could you explicitly call out that this supports Unix-style regex or `fnmatch` regex (as opposed to `re`)? Also, can we add an example using subclass matching?

aws_xray_sdk/ext/httplib/patch.py
```diff
@@ -22,8 +22,33 @@
...
+
+def reset_ignored():
+    global _XRAY_IGNORE
+    _XRAY_IGNORE.clear()
+    _ignored_add_default()
+
+
+def _ignored_add_default():
+    # skip httplib tracing for SDK built-in centralized sampling pollers
+    add_ignored(subclass='botocore.awsrequest.AWSHTTPConnection', urls=['/GetSamplingRules', '/SamplingTargets'])
```
willarmiros, line 47: I'm conflicted here - I know that using Python's built-in string matching as we are now will be faster than using `fnmatch`, but it would be awkward to not use this ignore mechanism and special-case /GetSamplingRules and /SamplingTargets. 

I guess the added latency is kinda peanuts compared to the actual network request, so it's probably ok, but what do you think @srprash?

willarmiros (commented):
Overall looks good, thanks for the contribution! Couple small comments.

aws_xray_sdk/ext/httplib/patch.py
srprash: I feel using `fnmatch` for ignoring /GetSamplingRules and /SamplingTarget is okay since it fits well in the overall mechanism to ignore any URL. I'm not really aware of the latency of `fnmatch` but my guess is that special casing the sampling urls and then matching the user urls would be roughly equivalent and won't cause much of a difference here.

srprash:
@jonathangreen `httplib`/`http.client` is a low-level module and most of the people would use `requests` or `urllib` for making HTTP calls. I think the `urllib` module uses the low-level `http.client` underneath but I'm not sure about `requests`. Do you think this ignore mechanism would work for someone using `requests` or `urllib` as well?

README.md
jonathangreen: @willarmiros updated to add both of those items.

jonathangreen:
@srprash I don't believe requests calles httplib under the hood, so this method likely won't work there.

It would be a nice future improvement though to add a similar method for ignoring calls that come from requests.

willarmiros (approved):
LGTM. We should expand this mechanism to other libs like `requests` but doesn't need to be in this PR probably.

srprash:
@jonathangreen 
Yeah, `requests` patch should have its own handling for this mechanism.
The `urllib` module is automatically patched if the `httplib` is patched since urllib builds upon httplib/http.client. I did a quick test with the following piece of code and it works! :)
```python
from aws_xray_sdk.core import patch
from aws_xray_sdk.ext.httplib import add_ignored

libraries = (['httplib'])
patch(libraries)

import urllib.request
add_ignored(hostname="www.amazon.com")
with urllib.request.urlopen('http://www.amazon.com') as response:
    html = response.read()
```

srprash (approved):
Looks good. Thanks!

Issue #124, opened 2019-01-24
Patching sqlalchemy


zhammer:
Is there a reason we can't just patch out the sqlalchemy `SessionMaker` with the `XRaySessionMaker`?

Curious why this is isolated from the rest of the patching.

chanchiem:
Hey Zhammer,

SQLAlchemy was written through the following pull request: https://github.com/aws/aws-xray-sdk-python/pull/14

And although it looks like it could've been written with the patcher, the current implementation utilized a different design specific to SQLAlchemy's base session and query classes.

zhammer:
I'm using postgres and actually just patched out `psycopg2` which has worked fine for me!

chanchiem:
Closing since you've found a solution to your underlying problem.

PR #264, opened 2020-12-17
Add support for SQLAlchemy Core

jonathangreen:
*Description of changes:*
This PR adds support for patching SQLAlchemy Core though the normal `patch` function. It can be patched by calling:
```python
patch(('sqlalchemy_core',))
```

This support can be used as a replacement, or to compliment the existing support for SQLAlchemy ORM. Because it patches at the core level, and all the ORM functions use core it will capture database communications either though the ORM or though core.


jonathangreen:
Since this uses the normal patching workflow its also a solution to #124.

willarmiros (approved):
Looks great to me! Thanks for the contribution!

Issue #225, opened 2020-05-11
Supported Python Libraries - Is psycopg2 supported?


tim-pugh:
one these site, it shows the supported python libs for x-ray: https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-python.html
https://docs.aws.amazon.com/xray-sdk-for-python/latest/reference/

But I found a PR for psycopg2 that seems to indicate it may be supported as well: https://github.com/aws/aws-xray-sdk-python/issues/80

looking at the source code, seems to me like it may be: https://github.com/aws/aws-xray-sdk-python/tree/master/tests/ext

Q1.) Is psycopg2 supported?
Q2.) Is there a system in place for updating docs? Maybe a tag that can used to assist keeping track of new additions?

bhautikpip:
Hi @tim-pugh ,

Thanks for opening this issue. Yes psycopg2 is supported. I will update this support in docs. In order to mitigate docs update issue in future I will create a new label like aws-docs-update to remind to update the docs as we merged the PR for new support features. Also, it will help tracking docs updates with PR. Thanks for the suggestion.

NathanielRN:
Closed in #266

PR #266, opened 2021-01-11
Update docs with supported libraries

srprash:
*Issue #, if available:* #225

The [aws docs](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-python-patching.html) seems to be updated with the current set of frameworks and libraries. Updating the [developer docs](https://docs.aws.amazon.com/xray-sdk-for-python/latest/reference/) as well to keep consistency.
Also updating tox.ini to use correct pymysql version since they dropped support for python 3.5 and below in v1.0.0.
 
### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -22,11 +22,16 @@ Currently supported web frameworks and libraries:
 * aioboto3/aiobotocore
 * aiohttp >=2.3
 * boto3/botocore
+* Bottle
 * Django >=1.10
 * Flask
-* httplib
+* httplib/http.client
 * mysql-connector
-* Pynamodb
+* pg8000
+* psycopg2
+* pymongo
+* pymysql
+* pynamodb
 * requests
 * SQLAlchemy
 * sqlite3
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -22,11 +22,16 @@ Currently supported web frameworks and libraries:
 * aioboto3/aiobotocore
 * aiohttp >=2.3
 * boto3/botocore
+* Bottle
 * Django >=1.10
 * Flask
-* httplib
+* httplib/http.client
```
bhautikpip, line 28: Should we use `,` here or put one in a new line? looks like `httplib` and `http.client` are two different client side libraries.

bhautikpip (approved):
Looks good. Just one small nit :)

docs/index.rst
srprash: httplib and http.client are both native libraries and are used for kind of same low level http clients. httplib is for python2 and http.client is for python3. So in a user's application, it will be either one or another and that's why I think "/" is fine.

Issue #154, opened 2019-05-03
TypeError when setting the key 'exceptions' for 'Entity.cause' 


cliffxuan:
If I understand correctly, `Entity.cause` is turned into a string from a dictionary, and raised a `TypeError` when calling `self.cause['exceptions'] = exceptions`.

Relevant traceback:
```
  File \\"/var/task/aws_xray_sdk/core/models/subsegment.py\\", line 31, in __call__\
    meta_processor=None,\
  File \\"/var/task/aws_xray_sdk/core/recorder.py\\", line 423, in record_subsegment\
    subsegment.add_exception(exception, stack)\
  File \\"/var/task/aws_xray_sdk/core/models/entity.py\\", line 228, in add_exception\
    self.cause[\'exceptions\'] = exceptions\
TypeError: \'str\' object does not support item assignment", "OrderId": "test8f2f5c5dbf774df6b900870bc50b2f7d", "LegacyOrderId": 20168, "Tenant": "uk", "requestId": "7cc85af8-a6a2-454f-8d63-b88339ada756", "XRayTraceId": "1-5ccbed5a-dff39c00f83f3190a93aa1f8"}
```

Full traceback:
```
b'_longitude=-3.2135957 (Caused by NewConnectionError(\'<urllib3.connection.HTTPConnection object at 0x7fc086466dd8>: Failed to establish a new connection: [Errno -2] Name or service not known\'))\
\
The above exception was the direct cause of the following exception:\
\
Traceback (most recent call last):\
  File \\"/var/task/aws_xray_sdk/core/recorder.py\\", line 401, in record_subsegment\
    return_value = wrapped(*args, **kwargs)\
  File \\"/var/task/lastmile/lambdas/predict/client.py\\", line 153, in predict_restaurant_working_time\
    response = self._call(url, params)\
  File \\"/var/task/lastmile/lambdas/predict/client.py\\", line 75, in _call\
    raise ConnectError(repr(exc)) from exc\
[...]

PR #267, opened 2021-01-13
Handling condition where Entity.cause is not a dict.

srprash:
*Issue* #154 

*Description of changes:*
In a case of exception chaining where an unhandled exception is captured in a subsegment and then propagated to the parent segment/subsegment, the `cause` property of this parent entity becomes a `str` from `dict` to record the exception id from the child subsegment. Any further dictionary operation this `cause` will result in TypeError.
Doing a type check on the Entity.cause when adding a new exception and resetting the cause to dict since we cannot recover the original cause dict once it has been updated to a string.

Also, making sure when adding multiple exceptions to an entity, they should be appended and not replaced.

### DIFF
--- a/tests/test_trace_entities.py
+++ b/tests/test_trace_entities.py
@@ -1,4 +1,6 @@
 # -*- coding: iso-8859-15 -*-
+from __future__ import unicode_literals
+
 import pytest
 
 from aws_xray_sdk.core.models.segment import Segment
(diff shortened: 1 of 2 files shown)


tests/test_trace_entities.py
```diff
@@ -4,6 +4,7 @@
 from aws_xray_sdk.core.models.segment import Segment
 from aws_xray_sdk.core.models.subsegment import Subsegment
 from aws_xray_sdk.core.models import http
+from aws_xray_sdk.core.models import throwable
```
willarmiros, line 7: Is this import used?

willarmiros (approved):
Thanks for the fix!

tests/test_trace_entities.py
srprash: removed

Issue #269, opened 2021-01-20
Xray not capturing stack traces in latest version of Flask


ianmetcalf:
The `after_request` handler is closing the segment before the `teardown_request` handler is called so the stack trace isn't captured.

It looks like this behavior changed as of Flask v1.1.0 (specifically https://github.com/pallets/flask/pull/3266)

Currently using:
- aws-xray-sdk v2.6.0
- flask v1.1.2

Here is our current work around:

```python
app = Flask(__name__)

def wrap_exception_handler(handler):
    def wrapped_function(exp):
        if not isinstance(exp, HTTPException) or exp.code >= 500:
            try:
                stack = stacktrace.get_stacktrace(
                    limit=xray_recorder.max_trace_back,
                )

                xray_recorder.current_segment().add_exception(exp, stack)
            except Exception:
                app.logger.error('Failed to add stack trace to xray segment')

        return handler(exp)
    return wrapped_function

app.handle_user_exception = wrap_exception_handler(app.handle_user_exception)
```

srprash:
Hi @ianmetcalf 
Thanks for identifying the issue and the root cause. You're right that the [`_handle_exception`](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/flask/middleware.py#L90) method which is the `teardown_request` handler is called after the segment is closed in `_after_request` and is [failing silently](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/flask/middleware.py#L99) while trying to add the exception. 
I believe there could be two options here. 
1. Do not close the segment in `_after_request` if the `response.status_code` is 500. Then the `_handle_exception` method will add the exception and close the segment.
2. Add a `handle_user_exception` handler method to the app similar to your workaround and add the exception there.

I'll test both the options and put out a fix soon.

srprash:
Did some digging into why the [unit tests](https://github.com/aws/aws-xray-sdk-python/blob/master/tests/ext/flask/test_flask.py#L91-L105) didn't catch this behavior change. It turns out that if we enable the testing mode, the `after_request` is not called in case of an unhandled exception which results in segment being still available in the `handle_exception` method and the test passes. 
[...]

PR #271, opened 2021-02-01
Close segment in only `_handle_exception` in case of Internal Server Error

srprash:
*Issue #, if available:* https://github.com/aws/aws-xray-sdk-python/issues/269

*Description of changes:*
- Return from `_after_request` when response code >= 500 so that the exception in recorded on the segment in the `_handle_exception` method.
-  Not enabling the testing mode so that the request is handled by the application in usual way during the exception

### DIFF
--- a/aws_xray_sdk/ext/flask/middleware.py
+++ b/aws_xray_sdk/ext/flask/middleware.py
@@ -81,6 +81,9 @@ class XRayMiddleware(object):
         if cont_len:
             segment.put_http_meta(http.CONTENT_LENGTH, int(cont_len))
 
+        if response.status_code >= 500:
+            return response
+
         if self.in_lambda_ctx:
             self._recorder.end_subsegment()
         else:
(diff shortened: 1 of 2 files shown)


willarmiros (approved):
LGTM

ianmetcalf:
@srprash I think this introduces a subtle bug. A 500 response in `after_request` doesn't always result in an exception being passed to `teardown_request`. For instance the following would cause the segment not to be closed at all.

```python
app.route('/error')
def error():
    return 'Server Error', 500
```

srprash:
@ianmetcalf Ah! I see. I was kind of unaware of tuple as a flask application response. I just found out about it [here](https://flask.palletsprojects.com/en/1.1.x/quickstart/#about-responses).
Good catch! I guess we should always close the segment in `teardown_request` if one is still present. I'll open a PR for the fix shortly.

PR #272, opened 2021-02-03
BugFix: Always close segment in teardown_request handler 

srprash:
*Issue #, if available:* https://github.com/aws/aws-xray-sdk-python/issues/269

*Description of changes:*
The previous PR which intended to fix the #269 introduced a [bug](https://github.com/aws/aws-xray-sdk-python/pull/271#issuecomment-772741466) where if the flask application does not pass an exception in case of response code >= 500, the segment will remain unclosed. This may be an edge case scenario that application faults do not pass exception to `teardown_request` handler but it is still a [valid case of responses](https://flask.palletsprojects.com/en/1.1.x/quickstart/#about-responses).
Thus, irrespective of exception being present or not, the middleware should always end the segment if its present. 

### DIFF
--- a/aws_xray_sdk/ext/flask/middleware.py
+++ b/aws_xray_sdk/ext/flask/middleware.py
@@ -17,7 +17,7 @@ class XRayMiddleware(object):
         self._recorder = recorder
         self.app.before_request(self._before_request)
         self.app.after_request(self._after_request)
-        self.app.teardown_request(self._handle_exception)
+        self.app.teardown_request(self._teardown_request)
         self.in_lambda_ctx = False
 
         if check_in_lambda() and type(self._recorder.context) == LambdaContext:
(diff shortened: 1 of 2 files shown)


srprash:
@ianmetcalf Feel free to take a look at this.

ianmetcalf:
@srprash This should take care of it.

With this change in place could you skip closing the segment in `after_request` for all requests. Having `teardown_request` close all segments would make things easier to reason about.

It might also be a good idea to rename the `_handle_exception` method since it's not only for exception handling now.

srprash:
@ianmetcalf That's a fair call out since the `teardown_request` is bound to be called at the end of each request, we can avoid some redundant code for closing the segment. And +1 for renaming the method.

willarmiros (approved):
LGTM

PR #275, opened 2021-03-17
Replace jsonpickle with json to serialize entity

lupengamzn:
*Issue #, if available:*

Security issue regarding `decode` API in `jsonpickle`: https://github.com/jsonpickle/jsonpickle/issues/335
X-Ray Python SDK doesn't use `decode` API, but better find a replacement

*Description of changes:*

Replace `jsonpickle` with `json` to serialize entity data

*Benchmark:*

Some initial benchmark results using `pytest-benchmark` on Python 2.7:

benchmark

Will automate the benchmark across the library shortly


aws_xray_sdk/core/models/entity.py
```diff
@@ -256,36 +257,42 @@ def get_origin_trace_header(self):
...
         Serialize to JSON document that can be accepted by the
-        X-Ray backend service. It uses jsonpickle to perform
-        serialization.
+        X-Ray backend service. It uses json to perform serialization.
         """
         try:
-            return jsonpickle.encode(self, unpicklable=False)
+            return json.dumps(self.to_dict(), default=str)
         except Exception:
             log.exception("got an exception during serialization")
```
srprash, line 265: We should log the exception as well to understand what went wrong. Can be at debug level.

aws_xray_sdk/core/utils/conversion.py
```diff
@@ -0,0 +1,35 @@
...
+        elif hasattr(obj, "__dict__"):
+            metadata = {}
+            for key, value in vars(obj).items():
+                if not callable(value) and not key.startswith('_'):
+                    metadata[key] = metadata_to_dict(value)
+            return metadata
+        else:
+            return obj
+    except Exception:
+        log.warning("Failed to convert " + str(obj) + " to dict")
```
srprash, line 34: Same. Let's log the exception at debug level.

tox.ini
```diff
@@ -18,8 +18,8 @@ deps =
     requests
     bottle >= 0.10
     flask >= 0.10
-    sqlalchemy
-    Flask-SQLAlchemy
+    sqlalchemy==1.3.*
```
srprash, line 21: I think we should only pin for py 3.4 and 3.5 and should still test with the latest version of sqlalchemy and flask-sqlalchemy for the supported python. Ideally, the SDK should work with every sqlalchemy version supported for each python version.

srprash (changes requested):
Looks good. Some minor comments.

tox.ini
lupengamzn: The other py SDKs will [fail](https://github.com/aws/aws-xray-sdk-python/runs/2143938554) the test with the latest sqlalchemy and flask-sqlalchemy version.

aws_xray_sdk/core/models/entity.py
lupengamzn: Updated

tox.ini
srprash: Why does the X-Ray SDK tests fail for the latest version of sqlalchemy? we are not using any of the internal APIs of sqlalchemy in our tests right? Does that mean the SDK is not compatible with latest sqlalchemy for supported python versions?

aws_xray_sdk/core/models/entity.py
srprash: I meant log the actual exception message and stack trace at debug level.

tox.ini
lupengamzn: Some of the unit tests are using the [API](https://github.com/aws/aws-xray-sdk-python/blob/master/tests/ext/sqlalchemy/test_query.py#L62)(for example), which will fail the test for the latest sqlalchemy and flask-sqlalchemy. Not sure if the Python SDK is compatible with the latest sqlalchemy or not, but at least some unit tests need to update the API they use from sqlalchemy and flask-sqlalchemy.

srprash (approved):
LGTM

aws_xray_sdk/core/models/entity.py
```diff
@@ -256,36 +257,42 @@ def get_origin_trace_header(self):
...
-        if not self.metadata:
-            del properties['metadata']
-        properties.pop(ORIGIN_TRACE_HEADER_ATTR_KEY, None)
-
-        del properties['sampled']
+        entity_dict = {}
+            
+        for key, value in vars(self).items():
+            if isinstance(value, bool) or value:
+                if key == 'subsegments':
```
bhautikpip, line 276: How do we handle the case if key is `annotations`, `aws` or `http` ? I see that you're handling `metadata` case separately. Shouldn't we do the same with `annotations` too? since in that case also we don't know how many key-value pairs would be there.

aws_xray_sdk/core/utils/conversion.py
```diff
@@ -0,0 +1,35 @@
...
+            metadata = {}
+            for key, value in vars(obj).items():
+                if not callable(value) and not key.startswith('_'):
+                    metadata[key] = metadata_to_dict(value)
+            return metadata
+        else:
+            return obj
+    except Exception:
+        log.exception("Failed to convert {} to dict".format(str(obj)))
+        return {}
```
bhautikpip, line 35: Probably adding a new line would be good.

bhautikpip (commented):
Should we also add performance results with this PR? This would help customers to understand the performance improvement in next release because of this serialization logic change.

aws_xray_sdk/core/models/entity.py
lupengamzn: `annotations`, `aws` or `http` are all `dict` with serializable values like `string`, `int` or `boolean`, so there is no need to handle these cases.

aws_xray_sdk/core/utils/conversion.py
lupengamzn: Added

lupengamzn:
> Should we also add performance results with this PR? This would help customers to understand the performance improvement in next release because of this serialization logic change.

We can do a follow-up PR to address this in future as seems like currently the `pytest-benchmark` library doesn't support all of the Python platforms supported by X-Ray Python SDK. So will need some workaround.

bhautikpip (approved):

aws_xray_sdk/core/models/entity.py
```diff
@@ -256,36 +257,42 @@ def get_origin_trace_header(self):
...
+                    # child subsegments are stored as List
+                    subsegments = []
+                    for subsegment in value:
+                        subsegments.append(subsegment.to_dict())
+                    entity_dict[key] = subsegments
+                elif key == 'cause':
+                    entity_dict[key] = {}
+                    entity_dict[key]['working_directory'] = self.cause['working_directory']
```
heitorlessa, line 284: @lupengamzn 

I couldn't dive into this implementation yet to provide a fix, but this is causing a regression in the AWS Lambda Powertools Tracer feature leading to a `TypeError: string must be a list of integers`

https://github.com/awslabs/aws-lambda-powertools-python/issues/383

PR #276, opened 2021-03-19
Add distribution channel smoke test

anuraaga:
*Issue #, if available:*

*Description of changes:*
Just builds without publishing any metrics for now

### DIFF
--- a/.github/workflows/continuous-monitoring.yml
+++ b/.github/workflows/continuous-monitoring.yml
@@ -1,0 +1,19 @@
+name: Continuous monitoring of distribution channels
+on:
+  workflow_dispatch:
+  schedule:
+    - cron:  '0 */6 * * *'
+
+jobs:
+  smoke-tests:
+    name: Run smoke tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - run: pip install tox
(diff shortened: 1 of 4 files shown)


.github/workflows/continuous-monitoring.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+      - run: pip install tox
+      - name: Run smoke tests
+        run: tox -c tox-distributioncheck.ini
```
srprash, line 19: How do we get notified if the workflow fails? Whom does GH sends notification to in case of a cron job?

anuraaga: I've sent a few similar PRs, but they're just establishing the build, notification is the next step. I think notification will be literally copy paste of the exact same code for all repos so can follow up with it.

bhautikpip: I think we can probably add like binary metric (`PythonSDKDistributionAvailabilityFailure`) which publishes value based on `Run smoke tests` is passed or not and then we can set alarms on that for notifications. I think that would be a good notification model. I agree with @anuraaga notifications can be next step.

bhautikpip (approved):
Looks good :)

srprash (approved):
Thanks!

Issue #281, opened 2021-03-25
AttributeError: 'function' object has no attribute '_yield_per' when upgrading to SQLAlchemy 1.4


mvanbaak:
```
Traceback (most recent call last):
  File "/Users/mvanbaak/dev/frontrow/public-api/src/fr_public_api/helpers/brandresolver.py", line 36, in resolve_brand_for_apikey
    key = db.session.query(ApiKey).filter_by(key=api_key).one()
  File "/Users/mvanbaak/.dotfiles/virtualenvs/fr_public_api/lib/python3.7/site-packages/aws_xray_sdk/ext/sqlalchemy/util/decorators.py", line 63, in wrapper
    res = func(*args, **kw)
  File "/Users/mvanbaak/.dotfiles/virtualenvs/fr_public_api/lib/python3.7/site-packages/sqlalchemy/orm/query.py", line 2730, in one
    return self._iter().one()
  File "/Users/mvanbaak/.dotfiles/virtualenvs/fr_public_api/lib/python3.7/site-packages/sqlalchemy/orm/query.py", line 2771, in _iter
    execution_options={"_sa_orm_load_options": self.load_options},
  File "/Users/mvanbaak/.dotfiles/virtualenvs/fr_public_api/lib/python3.7/site-packages/aws_xray_sdk/ext/sqlalchemy/util/decorators.py", line 63, in wrapper
    res = func(*args, **kw)
  File "/Users/mvanbaak/.dotfiles/virtualenvs/fr_public_api/lib/python3.7/site-packages/sqlalchemy/orm/session.py", line 1616, in execute
    _parent_execute_state is not None,
  File "/Users/mvanbaak/.dotfiles/virtualenvs/fr_public_api/lib/python3.7/site-packages/sqlalchemy/orm/context.py", line 253, in orm_pre_session_exec
    if "yield_per" in execution_options or load_options._yield_per:
AttributeError: 'function' object has no attribute '_yield_per'
```

Versions:
```
Flask==1.1.2
SQLAlchemy==1.4.2
Flask-SQLAlchemy==2.5.1

boto3==1.17.35
botocore==1.20.36
aws-xray-sdk==2.7.0
```

I get the same error with SQLAlchemy 1.4.0 and 1.4.1, things work fine with 1.3.23.

Confirmed it's an xray issue with the following change to my code which made things work again:
```
-db = XRayFlaskSqlAlchemy(model_class=Base)
+db = SQLAlchemy(model_class=Base)
[...]

PR #289, opened 2021-04-19
Fixing broken instrumentation for sqlalchemy >= 1.4.0

srprash:
*Issue #, if available:* #281

*Description of changes:*
For context, please read the following comments:
https://github.com/aws/aws-xray-sdk-python/issues/281#issuecomment-822618173
https://github.com/aws/aws-xray-sdk-python/pull/289#issuecomment-823461573
https://github.com/aws/aws-xray-sdk-python/pull/289#issuecomment-823512629

The changes being done in this PR are:
- Fixing the function decorator logic to check if the `obj` is of type function.
- Patching `Session.execute` method to capture subsegments for 2.0 style execution introduced in sqlalchemy v1.4.0. And some refactoring of the patch code.
- Adding a test case for the 2.0 style execution as per the [example](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#orm-query-is-internally-unified-with-select-update-delete-2-0-style-execution-available) in sqlalchemy docs. Since this test case would work only on v1.4.0 which is not available for py34 and py35, I added some test environments with conditions in `tox.ini` for the same.


srprash:
Looks like sqlalchemy 1.4.0 broke two functionalities in the x-ray SDK.
1. instrumentation of sqlalchemy using ORM
2. instrumentation of sqlalchemy_core using `patch` where the new ORM type statements are not getting patched by this [code](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/sqlalchemy_core/patch.py#L72).

srprash:
One thing we can do is since `Session.execute()` method is now the primary point of ORM statement execution, we can add a patch for it in the [`patch.py`](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/ext/sqlalchemy_core/patch.py#L72) with something like:
```python
wrapt.wrap_function_wrapper(
        'sqlalchemy.orm.session',
        'Session.execute',
        _xray_traced_sqlalchemy_session
    )
```

This seem to work fine and record a subsegment for `session.query()` and `session.add()` but it doesn't record the `sanitized_query` as plain sql statement in case of `session.add()` maybe due to the fact that the query is not unwrapped at this point in the execution.

ATM, I can say that for ORM style statement execution, it is safest to use the SQLAlchemy integration rather than patching `sqlalchemy_core` which was added in [this PR](https://github.com/aws/aws-xray-sdk-python/pull/264).

srprash:
I believe the reason for such a massive drop in the coverage is due to the fact that every test is run with `coverage run py.test ...` with different set of test cases. For each run of such command, only a single `.coverage` file is produced and overwriten by the run of next test cases.
Depending on which set of test cases are run as the command in the very last, the `.coverage` file (and thus the generated `coverage.xml`) will contain the coverage report about that run. In this case, I believe the last command to overwrite the `.coverage` was `coverage run --source aws_xray_sdk -m py.test tests/ext/sqlalchemy_core/ --ignore tests/ext/sqlalchemy_core/test_sqlalchemy_core_2.py` and so other modules' coverage dropped to 0% as per [this report](https://app.codecov.io/gh/aws/aws-xray-sdk-python/compare/289/tree/aws_xray_sdk/ext). 

This makes the coverage report quite unreliable, and the solution is to generate different coverage report for each test run and aggregate them together in a codecov report.

aws_xray_sdk/ext/sqlalchemy_core/patch.py
```diff
@@ -48,7 +52,15 @@ def _sql_meta(instance, args):
...
 def _xray_traced_sqlalchemy_execute(wrapped, instance, args, kwargs):
-    name, sql = _sql_meta(instance, args)
+    return _fetch_sql_metadata_and_process_request(wrapped, instance, args, kwargs)
+
+
+def _xray_traced_sqlalchemy_session(wrapped, instance, args, kwargs):
+    return _fetch_sql_metadata_and_process_request(wrapped, instance.bind, args, kwargs)
+
+
+def _fetch_sql_metadata_and_process_request(wrapped, engine_instance, args, kwargs):
```
anuraaga, line 62: What does the `fetch` mean? Is it fetching?

anuraaga (commented):
@srprash Is the coverage issue introduced by this PR or it already was there? The `tox.ini` change doesn't seem to be adding any new patterns

srprash:
> Is the coverage issue introduced by this PR or it already was there? The `tox.ini` change doesn't seem to be adding any new patterns

My suspicion is that the issue was always there and was sometimes encountered in PRs like [this](https://github.com/aws/aws-xray-sdk-python/pull/267) which saw decreased in coverage despite adding more tests.
I [experimented](https://github.com/srprash/aws-xray-sdk-python/pull/4) by rearranging the command order and saw the [coverage](https://github.com/srprash/aws-xray-sdk-python/pull/4#issuecomment-828145512) change from 79.42% to 31.98% between [this](https://github.com/srprash/aws-xray-sdk-python/pull/4/commits/4dee3bc08b7da9e1d084c57f876d92b771f6f1fa) and [this](https://github.com/srprash/aws-xray-sdk-python/pull/4/commits/8411c1b23efcd9bbc219399e05eb83aeb6218f45) commit.

aws_xray_sdk/ext/sqlalchemy_core/patch.py
srprash: By fetching if you mean getting info from some remote endpoint/service, then the function is not really _fetching_ any metadata rather it's just extracting info from db engine. Probably a better name can be `_populate_sql_metadata_and_process_request` or simply `_process_request`?

srprash: renamed

anuraaga (approved):
Cool in that case I don't think coverage issue blocks this PR, will be good to fix it separately at some point.

Issue #288, opened 2021-04-11
Error when serializing segment causes invalid message to be sent to the daemon


trossyvormes:
#283 makes `Entity.to_dict` throw an exception in some cases. When `to_dict` throws an exception, `Entity.serialize` returns `None`:

https://github.com/aws/aws-xray-sdk-python/blob/7602447db1b5dbd98fcda804942e86853cb50809/aws_xray_sdk/core/models/entity.py#L264-L265

Anyway, the `None` is put into the message that is sent to the daemon, so the daemon gets `{"format":"json","version":1}\nNone`, which is an invalid message.

https://github.com/aws/aws-xray-sdk-python/blob/7602447db1b5dbd98fcda804942e86853cb50809/aws_xray_sdk/core/emitters/udp_emitter.py#L35-L37

I don't think catching everything and returning `None` in `Entity.serialize` is a good idea. The exception should continue to be raised and the message send request should be aborted.

See also aws/aws-xray-daemon#120.

lupengamzn:
Hey @trossyvormes ,

Thank you for pointing this out

We're fixing the Python serialization issue in [PR](https://github.com/aws/aws-xray-sdk-python/pull/284) and as for the Daemon issue you mentioned, we figured out that this is an issue introduced in the latest release happened last Friday. We'll do a patch release for both Python SDK and Daemon soon.

Sorry for the inconvenience this brought to you.

trossyvormes:
> Hey @trossyvormes ,
> 
> Thank you for pointing this out
> 
> We're fixing the Python serialization issue in [PR](https://github.com/aws/aws-xray-sdk-python/pull/284) and as for the Daemon issue you mentioned, we figured out that this is an issue introduced in the latest release happened last Friday. We'll do a patch release for both Python SDK and Daemon soon.
> 
> Sorry for the inconvenience this brought to you.

Hello Lu. Fixing the serialization error is a band-aid to the real problem which is that `None` is returned by `Entity.serialize` when `to_dict` fails. If there is a new bug in `Entity.to_dict` that raises an exception you will get the same bug again. `Entity.serialize` has to change and not return `None` when `to_dict` fails.

lupengamzn:
@trossyvormes Yes, totally agree that SDK should do the sanity check for this case. Will do a followup PR.

lupengamzn:
Closing this issue as it's addressed in PR #292

PR #292, opened 2021-04-26
Handle exception when sending entity to Daemon

lupengamzn:
*Issue #, if available:*

https://github.com/aws/aws-xray-sdk-python/issues/288

*Description of changes:*

- Discard sending entity to Daemon if any exception happens during serialization

### DIFF
--- a/aws_xray_sdk/core/emitters/udp_emitter.py
+++ b/aws_xray_sdk/core/emitters/udp_emitter.py
@@ -60,12 +60,7 @@ class UDPEmitter(object):
         return self._port
 
     def _send_data(self, data):
-
-        try:
-            self._socket.sendto(data.encode('utf-8'), (self._ip,
-                                self._port))
-        except Exception:
-            raise
+        self._socket.sendto(data.encode('utf-8'), (self._ip, self._port))
 
     def _parse_address(self, daemon_address):
         try:
--- a/aws_xray_sdk/core/emitters/udp_emitter.py
+++ b/aws_xray_sdk/core/emitters/udp_emitter.py
@@ -32,12 +32,15 @@ class UDPEmitter(object):
 
         :param entity: a trace entity to send to the X-Ray daemon
         """
-        message = "%s%s%s" % (PROTOCOL_HEADER,
-                              PROTOCOL_DELIMITER,
-                              entity.serialize())
+        try:
+            message = "%s%s%s" % (PROTOCOL_HEADER,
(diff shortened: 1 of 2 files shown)


aws_xray_sdk/core/emitters/udp_emitter.py
```diff
@@ -32,12 +32,15 @@ def send_entity(self, entity):
 
         :param entity: a trace entity to send to the X-Ray daemon
         """
-        message = "%s%s%s" % (PROTOCOL_HEADER,
-                              PROTOCOL_DELIMITER,
-                              entity.serialize())
+        try:
```
srprash, line 35: The `try-except` block should be tighter if we really just want to skip sending data to daemon in case of serialization failure. With the current try-except block and the log message, we won't know if the exception was in entity serialization or in sending data to daemon.

lupengamzn: The `log.exception` will print out the trace stack so we can know where it goes wrong, either during serialization or sending data to daemon.

srprash: Oh yes. Right.

aws_xray_sdk/core/emitters/udp_emitter.py
```diff
@@ -62,7 +65,7 @@ def _send_data(self, data):
             self._socket.sendto(data.encode('utf-8'), (self._ip,
                                 self._port))
         except Exception:
-            log.exception('failed to send data to X-Ray daemon.')
+            raise
```
srprash, line 68: Do we need to explicitly `raise` the exception here if we are not specially handling it? I think we can remove the `try-except` block here.

aws_xray_sdk/core/models/entity.py
```diff
@@ -262,7 +262,7 @@ def serialize(self):
         try:
             return json.dumps(self.to_dict(), default=str)
         except Exception:
-            log.exception("Failed to serialize %s", self.name)
+            raise
```
srprash, line 265: Same here. We don't need `try-except` clock if we are just raising exception. I think it would be better to log "Failed to serialize entity" message in the `except` and then raise the exception.

lupengamzn: I removed the `try-catch` block and don't think we need to log the message as the detailed stack trace will be printed out anyway

srprash (approved):

PR #293, opened 2021-04-27
no op trace id generation

bhautikpip:
*Issue #, if available:*

*Description of changes:*

By merging this PR X-Ray Python SDK will generate no op trace and entity id for upsampled requests and secure random id for sample requests. However, customer can choose to generate secure random trace and entity ids for all the requests by setting `AWS_XRAY_NOOP_ID` environment variable as `False`.


README.md
```diff
@@ -192,6 +192,9 @@ if xray_recorder.is_sampled():
     xray_recorder.put_metadata('mykey', metadata)
 ```
 
+### Generate NoOp Trace and Entity Id
+X-Ray Python SDK will by default generate no-op trace and entity id for unsampled requests and secure random trace and entity id for sampled requests. If customer wants to enable generating secure random trace and entity id for all the (sampled/unsampled) requests (this is applicable for trace id injection into logs use case) then they should set the `AWS_XRAY_NOOP_ID` envrionment variable as False.
```
willarmiros, line 196:
```suggestion
X-Ray Python SDK will by default generate no-op trace and entity id for unsampled requests and secure random trace and entity id for sampled requests. If customer wants to enable generating secure random trace and entity id for all the (sampled/unsampled) requests (this is applicable for trace id injection into logs use case) then they should set the `AWS_XRAY_NOOP_ID` environment variable as False.
```

aws_xray_sdk/core/models/dummy_entities.py
```diff
@@ -12,9 +14,13 @@ class DummySegment(Segment):
     A dummy segment will not be sent to the X-Ray daemon. Manually create
     dummy segments is not recommended.
     """
-    def __init__(self, name='dummy'):
 
-        super(DummySegment, self).__init__(name=name, traceid=TraceId().to_id())
+    def __init__(self, name='dummy'):
+        noOpID = os.getenv('AWS_XRAY_NOOP_ID')
```
willarmiros, line 19: nit: snake case (e.g. `no_op_id`)

aws_xray_sdk/core/models/dummy_entities.py
```diff
@@ -12,9 +14,13 @@ class DummySegment(Segment):
...
+        if not noOpID or noOpID.lower() == 'true':
```
willarmiros, line 20: We should probably explicitly check for the `false` case, and generate a random ID if and only if the environment variable is set to false, since that's what our docs say. Looks like that's how we did it in Go too: https://github.com/aws/aws-xray-sdk-go/pull/293

aws_xray_sdk/core/models/dummy_entities.py
```diff
@@ -79,9 +85,12 @@ class DummySubsegment(Subsegment):
     to a dummy subsegment becomes no-op. Dummy subsegment will not
     be sent to the X-Ray daemon.
     """
-    def __init__(self, segment, name='dummy'):
 
+    def __init__(self, segment, name='dummy'):
         super(DummySubsegment, self).__init__(name, 'dummy', segment)
+        noOpID = os.getenv('AWS_XRAY_NOOP_ID')
+        if not noOpID or noOpID.lower() == 'true':
```
willarmiros, line 92: Same here for checking for `false` instead of `true`. Also, don't we need an else statement? Otherwise the original subsegment ID is always generated.

tests/test_traceid.py
```diff
@@ -9,3 +12,75 @@ def test_id_format():
...
+    os.environ['AWS_XRAY_NOOP_ID'] = 'True'
+    segment = xray_recorder.begin_segment('segment_name', sampling=False)
+
+    # Start and end a subsegment
+    subsegment = xray_recorder.begin_subsegment('subsegment_name')
+    xray_recorder.end_subsegment()
+
+    # Close the segment
+    xray_recorder.end_segment()
+    os.unsetenv('AWS_XRAY_NOOP_ID')
```
willarmiros, line 61: Can we unset the environment variable in a separate cleanup block? This way if one of these tests unexpectedly fails we won't have stale environment variables lying around. Example: https://github.com/aws/aws-xray-sdk-python/blob/d025f5641496efdc9dfad2b270139c224053885c/tests/test_sdk_config.py#L9-L10

willarmiros (commented):
Mostly minor callouts

aws_xray_sdk/core/models/dummy_entities.py
bhautikpip: This [line](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/subsegment.py#L98) calls the entity object and generates the random entity id (segment id). So basically that is why I have added env var checking condition in `DummySubSegment` class so that it initializes entity id (segment id) with `000000000`. So basically when env var value is `true` or not set which is the default case we generate entity id with `000000000` otherwise this line https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/subsegment.py#L98 will generate secure random entity id. Also, this condition https://github.com/aws/aws-xray-sdk-python/pull/293/files#diff-18ac8509f2bfc4403d165a3517044c0dec95f4c1f52a002fc4f6564cfcf351bdR31  will help us achieve that. So probably that is why we don't need checking for `false` condition.

aws_xray_sdk/core/models/dummy_entities.py
bhautikpip: Here we can explicitly check for the `false` case but in the subsegment case we are checking for `true` so I have kept it here like this.

aws_xray_sdk/core/models/dummy_entities.py
bhautikpip: There will be 3 cases. 1, env var value `true` 2, env var value `false` 3, default. So, case 1 and 3 covered in this condition and for case 2 only we would generate secure random id.

willarmiros (approved):
LGTM, thanks for changes!

Issue #305, opened 2021-06-15
'import pkg_resources' is slow and impacts cold starts


prestomation:
There are two locations in the code base that use `from pkg_resources import resource_filename` to get the location of some packaged, static json data.

During some profiling of a ~1.5gb python3.8 lambda function, I discovered that this import alone takes on the order of 250ms and is one of the larger single contributors of cold start in my sample function.

Could we consider some other method for consuming this data?

NathanielRN:
Thanks for bringing this to our attention, @prestomation !

I was able to reproduce the slow behavior of `resource_filename` and found a solution online to make the loading faster. We just merged #306 in hopes of improving the performance, which should go out in the next release.

Let me know if you have any additional questions or comments!

heitorlessa:
Hey @NathanielRN do we have some rough dates on when the next release will be out?

It's been a couple of months since the last release. It'd be awesome to have this Perf improvement out for customers

NathanielRN:
Thanks for the prompt @heitorlessa ! I'll message the team and see if they can cut us a release soon.

NathanielRN:
This should have been released in be4dea2f89e44dd52c68b23e1620b8541947e64b as version `2.9.0`. Closing this issue.

PR #306, opened 2021-06-28
Replace slow json filename resolver

NathanielRN:
*Issue #305* 

*Description of changes:*

For optimization purposes, we found that `pkgutil.get_data(package, resource)` method could be used to read a file in a much faster way than `from pkg_resources import resource_filename`.

Learn more about [reading static files in Python](https://stackoverflow.com/questions/6028000/how-to-read-a-static-file-from-inside-a-python-package#comment108239645_58941536).

*How did I test*

I used `pytest-benchmark` to compare the performance of the two:

```
$ pytest test_local_sampling_benchmark.py
======================================================================================== test session starts =========================================================================================
platform darwin -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
benchmark: 3.4.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /Users/enowell/git/aws-xray-sdk-python
plugins: benchmark-3.4.1
collected 2 items

test_local_sampling_benchmark.py ..                                                                                                                                                            [100%]


---------------------------------------------------------------------------------------------- benchmark: 2 tests ---------------------------------------------------------------------------------------------
Name (time in us)                      Min                    Max                Mean              StdDev             Median                IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_pkgutil_static_read           50.6210 (1.0)         828.0920 (1.0)       61.8523 (1.0)       22.4959 (1.0)      54.3770 (1.0)      10.4147 (1.0)       798;992       16.1676 (1.0)       11205           1
test_pkg_resources_static_read     64.9590 (1.28)     18,941.0130 (22.87)    100.6223 (1.63)     558.5882 (24.83)    79.4150 (1.46)     12.1315 (1.16)        2;111        9.9382 (0.61)       1144           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
========================================================================================= 2 passed in 2.38s ==========================================================================================
```

which shows that `pkgutil` is faster.

*Noteworthy points*

* `pkgutil` is standard in `python2` as well, so it should work fine with it.
* I had a chance to do an e2e test and traces still make it to AWS X-Ray

### DIFF
--- a/tests/test_local_sampling_benchmark.py
+++ b/tests/test_local_sampling_benchmark.py
@@ -1,0 +1,16 @@
+import json
+import pkgutil
+from pkg_resources import resource_filename
+
+from aws_xray_sdk.core.sampling.local.sampler import LocalSampler
+
+def test_pkg_resources_static_read(benchmark):
+    def get_sampling_rule():
+        with open(resource_filename(__name__, 'mock_sampling_rule.json')) as f:
+            return json.load(f)
+    benchmark(get_sampling_rule)
+
+def test_pkgutil_static_read(benchmark):
+    def get_sampling_rule():
+        json.loads(pkgutil.get_data(__name__, 'mock_sampling_rule.json'))
+    benchmark(get_sampling_rule)
--- a/tests/test_local_sampling_benchmark.py
+++ b/tests/test_local_sampling_benchmark.py
@@ -2,15 +2,15 @@ import json
 import pkgutil
 from pkg_resources import resource_filename
 
-from aws_xray_sdk.core.sampling.local.sampler import LocalSampler
+# Faster
+def test_pkgutil_static_read(benchmark):
+    def get_sampling_rule():
+        return json.loads(pkgutil.get_data(__name__, 'mock_sampling_rule.json'))
+    benchmark(get_sampling_rule)
 
+# Slower
 def test_pkg_resources_static_read(benchmark):
     def get_sampling_rule():
         with open(resource_filename(__name__, 'mock_sampling_rule.json')) as f:
             return json.load(f)
-    benchmark(get_sampling_rule)
-
-def test_pkgutil_static_read(benchmark):
-    def get_sampling_rule():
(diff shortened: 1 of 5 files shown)


anuraaga (approved):
Nice find

anuraaga:
Looks like this repo's CI needs to be migrated to github

tests/test_local_sampling_benchmark.py
```diff
@@ -0,0 +1,16 @@
...
+def test_pkg_resources_static_read(benchmark):
+    def get_sampling_rule():
+        with open(resource_filename(__name__, 'mock_sampling_rule.json')) as f:
+            return json.load(f)
+    benchmark(get_sampling_rule)
+
+def test_pkgutil_static_read(benchmark):
+    def get_sampling_rule():
+        json.loads(pkgutil.get_data(__name__, 'mock_sampling_rule.json'))
+    benchmark(get_sampling_rule)
```
NathanielRN, line 16: These tests are really testing the performance of standard library functions (`pkgutil` and `pkg_resources`) instead of testing `aws-xray-sdk-python` functions.

For that reason I think we should not include them here? The PR comment can serve as proof that this change is worth it.

Otherwise, we can leave the tests here so the results can be reproduced later.

anuraaga: Agree with deleting these benchmarks (first thing that came to mind but keeping them also isn't a big deal)

srprash (approved):

PR #314, opened 2021-10-19
Working docker demo on python 3.9 with the AWS X-Ray daemon installed in the docker image. also includes a script that fires off many API (curl) requests that are used to verify your AWS X-Ray IAMs are correct. 

jay-johnson:
*Issue #, if available:*

The master branch failed to build the Docker image on python 3.6 due to urllib3 pip versioning conflicts.

*Description of changes:*

#### Included build works on python 3.9 with the docker image using:

1. Runs as a non-root user (``driver``)
2. A python virtualenv
3. AWS CLI v2 installed
4. AWS X-Ray daemon installed (enabled/disabled on startup with the environment variable ``XRAY_DAEMON_ENABLED``)
 
#### Added a docker build script, a docker compose start script, and a readme showing how to:

1. Set up the demo (``build.sh`` + ``start.sh``)
2. Run the demo script that fires off curl tests (``run-xray-demo.sh``)
3. Debug issues

#### General fixes

1. Demo uses docker compose with AWS environment variable installation to verify IAM access
2. Removed pinned python pip versions in the requirements.txt
3. Added flake8/pep8 
4. Colorized logging
5. Support for additional logging using an environment variable ``VERBOSE=1`` on container startup

Assuming the flask application and the AWS X-Ray daemon are running correctly, the test script's API calls will show up in the AWS X-Ray UI for review.


NathanielRN:
Thanks for this change! I can confirm that the Docker Sample App does not work on master:

```
$ docker build -t aws-xray-python-demo-app .
…
#9 5.161 ERROR: Cannot install -r requirements.txt (line 10) and urllib3==1.26.5 because these package versions have conflicting dependencies.
#9 5.161
#9 5.161 The conflict is caused by:
#9 5.161     The user requested urllib3==1.26.5
#9 5.161     requests 2.23.0 depends on urllib3!=1.25.0, !=1.25.1, <1.26 and >=1.21.1
#9 5.161
#9 5.161 To fix this you could try to:
#9 5.161 1. loosen the range of package versions you've specified
#9 5.161 2. remove package versions to allow pip attempt to solve the dependency conflict
#9 5.161
#9 5.161 ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
#9 5.348 WARNING: You are using pip version 21.2.4; however, version 21.3.1 is available.
#9 5.348 You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
------
executor failed running [/bin/sh -c pip install -r requirements.txt]: exit code: 1
```

And I can see that building it using your branch fixes the problem and the app runs well 🙂

I really enjoyed reading your PR because your bash scripts are really useful and I learned a lot of neat things about them! However, I'm thinking we shouldn't take them on in this repo at this time because the goal is to just show a quick demo of how you might use the AWS X-Ray SDK and these scripts and extra packages would increase our technical debt quite a bit 😕

I'm not sure why it was pinned to that version originally, but I noticed that if we change

```python
urllib3==1.26.5
```

to

```python
urllib3~=1.25
```

then the Dockerfile works for both `python3.6` and `python3.9`! If that's okay with you, would you be willing to make this PR only have this one line change? 🙂 It's a great call out and hopefully this way we can make the experience better for new users.

One other thing that I think your PR does great is that it adds a README to this folder! We could definitely benefit from that, but perhaps we could just mention using the regular `docker build -t aws-xray-python-demo-app .` command just to help get users started?

Let me know what you think!

NathanielRN (commented):
Just some passing comments I saw on `.gitignore` and the `Dockerfile`, let me know if you have any questions? 🙂

PR #320, opened 2021-12-18
Drop PyPI mock package dependency on Python 3.3+

musicinmybrain:
Use `unittest.mock` from the standard library where available.

*Issue #, if available:* *N/A*

*Description of changes:*

Conditionally require the PyPI `mock` backport package in `tox.ini`: instead of always, only for Python versions older than 3.3.

Wherever `mock` or `unittest.mock` is imported, support either.

See https://fedoraproject.org/wiki/Changes/DeprecatePythonMock for context.

### DIFF
--- a/tests/ext/aiohttp/test_middleware.py
+++ b/tests/ext/aiohttp/test_middleware.py
@@ -5,7 +5,10 @@ Expects pytest-aiohttp
 """
 import asyncio
 from aws_xray_sdk import global_sdk_config
-from unittest.mock import patch
+try:
+    from unittest.mock import patch
+except ImportError:
+    from mock import patch  # Python 2
 
 from aiohttp import web
 from aiohttp.web_exceptions import HTTPUnauthorized
--- a/tests/ext/aiohttp/test_middleware.py
+++ b/tests/ext/aiohttp/test_middleware.py
@@ -8,7 +8,8 @@ from aws_xray_sdk import global_sdk_config
 try:
     from unittest.mock import patch
 except ImportError:
-    from mock import patch  # Python 2
+    # NOTE: Python 2 dependency
+    from mock import patch
 
 from aiohttp import web
 from aiohttp.web_exceptions import HTTPUnauthorized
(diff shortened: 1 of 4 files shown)


srprash (approved):
LGTM! 
@NathanielRN Can you take a look as well? Thanks!

tox.ini
```diff
@@ -36,7 +36,7 @@ deps =
     testing.postgresql
     testing.mysqld
     webtest
-    mock
+    mock ; python_version < '3.3'
 
     # Python2 only deps
     py{27}: enum34
```
NathanielRN, line 42: Maybe we can move the dependency to the `Python2 only deps` section to be consistent?

```suggestion
    # Python2 only deps
    py{27}: enum34, mock
```

tests/ext/aiohttp/test_middleware.py
```diff
@@ -5,7 +5,10 @@
 """
 import asyncio
 from aws_xray_sdk import global_sdk_config
-from unittest.mock import patch
+try:
+    from unittest.mock import patch
+except ImportError:
+    from mock import patch  # Python 2
```
NathanielRN, line 11: I like putting the comments above but this is just an opinion. It's fine the way it is too.

```suggestion
except ImportError:
    # NOTE: Python 2 dependency
    from mock import patch
```

NathanielRN (approved):
Thanks for this change! Left one comment 😄

tests/ext/aiohttp/test_middleware.py
musicinmybrain: Adjusted in the three files where this sort of inline comment appears.

musicinmybrain:
Applied @NathanielRN’s suggestions.

tox.ini
```diff
@@ -36,10 +36,8 @@ deps =
     testing.postgresql
     testing.mysqld
     webtest
-    mock
-
     # Python2 only deps
-    py{27}: enum34
+    py{27}: enum34, mock
```
NathanielRN, line 40: Sorry my bad I think there should not be a comma for pip dependencies, committing this change:

```suggestion
    py{27}: enum34 mock
```

Issue #321, opened 2021-12-19
Support Python 3.10 (removed asyncio loop keyword)


musicinmybrain:
It would be nice to have official support for Python 3.10.

The main issue I see is that the `loop` parameter in the high-level `asyncio` APIs, deprecated since 3.8, is [removed in 3.10](https://docs.python.org/3/whatsnew/3.10.html#whatsnew310-removed). These are still used in quite a few places.

It’s not easy for me to run the *full* test suite, so it’s hard for me to know whether or not simply removing all such `loop` parameters without further changes to the code would fix Python 3.10 support without breaking anything on some older Python version that you still support.

srprash:
Hi @musicinmybrain 
Thanks for bringing this to our notice. I see that the potential impact of this API change could be in the [`async_context`](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/async_context.py) which uses `asyncio`.
We will have the change tested on our end soon to address the compatibility with Python 3.10.

yan12125:
> I see that the potential impact of this API change could be in the [async_context](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/async_context.py) which uses `asyncio`.

Maybe an API change is not necessary. In async_context.py, `loop` is passed to `asyncio.current_task()`, which still accepts a loop argument as of Python 3.10.

musicinmybrain:
Python 3.10 and 3.11 seem to be properly supported as of [2.12.0](https://github.com/aws/aws-xray-sdk-python/blob/2.12.0/CHANGELOG.rst#2120), so I’m closing this issue.

PR #333, opened 2022-04-21
Remove distutils.util.strtobool deprecated in PEP 632 (Python 3.10)

ScriptSmith:
*Description of changes:*
[PEP 632](https://peps.python.org/pep-0632/) deprecates `distutils` in Python 3.10, and recommends specifically that `distutils.util.strtobool` be replaced by reimplementing it, which I have done here to preserve existing behaviour.

Related to: #321


NathanielRN (approved):
Thanks for this change! 🚢

Issue #310, opened 2021-07-29
concurrent asyncio tasks and xray are not working with Lambda


abivolmv:
Hello,
I am opening this issue on request of @NathanielRN after a discussion with him in https://github.com/aws/aws-xray-sdk-python/issues/203 .

The problem is that when using aioboto3, asyncio (with gather or wait) and xray in Lambda we either get an exception or no subsegments showing calls to S3 or SQS.
Here is the last code snippet from that issue:

```python
import asyncio
from io import BytesIO

import aioboto3
from aws_xray_sdk.core import xray_recorder, patch_all
from aws_xray_sdk.core.async_context import AsyncContext

# here I tried adding your solution (not added when solution to use AsyncContext() is used below)
# xray_recorder.configure(service='repro_xray_issue')  # uncomment this for second solution trial

patch_all()


def lambda_handler(a,b):
    asyncio.run(main())


async def main():
    # here I tried adding your solution (not added when solution to remove AsyncContext() is used above)
    xray_recorder.configure(service='repro_xray_issue', context=AsyncContext())    # comment this for second solution trial
    async with xray_recorder.in_segment_async('my_segment_name') as segment:  # comment this for second solution trial
        filelike1 = BytesIO()
        filelike2 = BytesIO()
        res1, res2 = await asyncio.gather(s3_get(filelike1), s3_get(filelike2))  # .wait() doesn't work either


@xray_recorder.capture_async('s3_get')  # tried with this and without also
async def s3_get(filelike):
    async with aioboto3.Session().client('s3') as s3:
        return await s3.download_fileobj('s3-validation-files-003', 'test.txt', filelike)
```

willarmiros:
Hi @abivolmv,

Thanks for creating this separate issue and posting some reproduction code. We have this as an investigation item in our backend and will follow up when we have bandwidth.

NathanielRN:
Adding some notes I took down during my quick investigation of this issue:

It looks like our `AsyncContextManager` is able to `__aexit__` before the `s3` tasks initiated by `asyncio` finish. Maybe it's not put on the `taskLoop` correctly OR it just finishes faster so that they finish before the other tasks finish creating and ending their sub segments. A reminder that this bug sometimes happens and sometimes does not happen.

I'm referring lines specifically to these lines of the `async_recorder.py` for `AsyncContextManager`:

[...]

Issue #164, opened 2019-07-01
async operations with patched aiobotocore gives error "Already ended segment and subsegment cannot be modified."


IvDoorn:
I'm running into some problems with XRay in combination with async functions and aiobotocore.

Packages:
```
aioboto3==6.4.1
aiobotocore==0.10.2
aws-xray-sdk==2.4.2
boto3==1.9.176
botocore==1.12.176
```

I've created a very small sample script to demonstrate my issue:
```
import aioboto3
import asyncio
import logging
import sys

from datetime import datetime

from aws_xray_sdk.core import patch
from aws_xray_sdk.core import xray_recorder
from aws_xray_sdk.core.async_context import AsyncContext

logger = logging.getLogger()
logger.setLevel(logging.INFO)

xray_logger = logging.getLogger('aws_xray_sdk')
xray_logger.setLevel(logging.INFO)

handler = logging.StreamHandler(sys.stdout)
handler.setLevel(logging.DEBUG)
logging.getLogger().addHandler(handler)

xray_recorder.configure(service='Test', context=AsyncContext(), sampling=False)

patch([
    'aiobotocore',
], raise_errors=False)


async def perform_test(gather=False, loop_count=10):
    async_client = aioboto3.client('clouddirectory', region_name='eu-west-1')

    try:
        with xray_recorder.in_segment_async('ASYNC'):
            @xray_recorder.capture_async('execute')
            async def execute(client):
                retval = await client.list_directories(state='ENABLED')
                return retval['Directories']

            results = []
            if gather:
                with xray_recorder.capture_async('gather'):
                    for result in await asyncio.gather(*[execute(async_client) for i in range(loop_count)]):
                        results += result

            else:
                with xray_recorder.capture_async('loop'):
                    for i in range(loop_count):
                        results += await execute(async_client)

            return results
    finally:
        await async_client.close()


def test_gather(loop_count=10):
    try:
        now = datetime.utcnow()
        asyncio.get_event_loop().run_until_complete(perform_test(gather=True, loop_count=loop_count))
        logger.info('GATHER: PASS %s' % str(datetime.utcnow() - now))
    except:
        logger.exception('GATHER: FAIL')


def test_loop(loop_count=10):
    try:
        now = datetime.utcnow()
        asyncio.get_event_loop().run_until_complete(perform_test(gather=False, loop_count=loop_count))
        logger.info('LOOP: PASS %s' % str(datetime.utcnow() - now))
    except:
[...]

PR #340, opened 2022-06-09
Bug Fix: Defensively copy context entities

tyler-dodge:
# Description

Before this change, concurrent async tasks would all share the same instance of the entities list. This change makes it so they each get their own copy of the list.

This matters because the recorder modifies the `entities` list in place, which makes it so concurrent subtasks end up looking at the wrong item in the `entites` list when deciding the parent subsegment:

https://github.com/aws/aws-xray-sdk-python/blob/0f13101e4dba7b5c735371cb922f727b1d9f46d8/aws_xray_sdk/core/context.py#L90-L101

You can see this in the unit test added. Before this change the unit test would print out the following:

```
Subseg 1 ID: 5b1c1cf70db2e808
Subseg 2 ID: 4a138f09cd3c6813
Subseg 3 ID: f97c02073584be8a
Subseg 4 ID: efac5645672e1cd6
Subseg 5 ID: 6874f083c496d388
Subseg 6 ID: f447815d60c734d6
Subseg 7 ID: edf7d60ea3d6875e
Subseg 8 ID: d2a1980e83a569f4
Subseg 9 ID: 6a0f1059ca4a91b6
Subseg 10 ID: 75421c2a04106214

Parent Segment ID: a9037c7ccc92e0c8 # Correct!
Subseg parent ID: a9037c7ccc92e0c8 # Correct!
Subseg parent ID: 4a138f09cd3c6813 # WRONG (all the ones below are wrong)
Subseg parent ID: f97c02073584be8a
Subseg parent ID: efac5645672e1cd6
Subseg parent ID: 6874f083c496d388
Subseg parent ID: f447815d60c734d6
Subseg parent ID: edf7d60ea3d6875e
Subseg parent ID: d2a1980e83a569f4
Subseg parent ID: 6a0f1059ca4a91b6
Subseg parent ID: 75421c2a04106214
```

With this change, all the subsegments have the correct parent ID.

(enowell) You can also visualize this in X-Ray from my tests:

Before:

image

Subsegments are children of each other even though they are concurrent:

After:

image

Subsegments have the correct parent ID.

~I think this possibly fixes the issue here:~

Fixes: #310
Fixes: #164

### DIFF
--- a/tests/test_async_recorder.py
+++ b/tests/test_async_recorder.py
@@ -3,6 +3,7 @@ import platform
 from .util import get_new_stubbed_recorder
 from aws_xray_sdk.version import VERSION
 from aws_xray_sdk.core.async_context import AsyncContext
+import asyncio
 
 
 xray_recorder = get_new_stubbed_recorder()
@@ -42,6 +43,28 @@ async def test_capture(loop):
     service = segment.service
     assert platform.python_implementation() == service.get('runtime')
     assert platform.python_version() == service.get('runtime_version')
+
+async def test_concurrent_calls(loop):
+    xray_recorder.configure(service='test', sampling=False, context=AsyncContext(loop=loop))
+    async with xray_recorder.in_segment_async('segment') as segment:
+        global counter
+        counter = 0
+        total_tasks = 10
+        event = asyncio.Event()
+        async def assert_task():
+            async with xray_recorder.in_subsegment_async('segment') as subsegment:
+                global counter
+                counter += 1
+                # Ensure that the task subsegments overlap
+                if counter < total_tasks:
+                    await event.wait()
+                else:
+                    event.set()
+                return subsegment.parent_id
+        tasks = [assert_task() for task in range(total_tasks)]
+        results = await asyncio.gather(*tasks)
+        for result in results:
+            assert result == segment.id
 
 
 async def test_async_context_managers(loop):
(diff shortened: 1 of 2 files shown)


NathanielRN:
Thanks @tyler-dodge! Just curious, do you have a screenshot of this failing or the bug you observed? I think this change makes sense but I think it would be useful to see it breaking. (Or maybe unit tests that show it works if that’s easier to add?).

tyler-dodge:
I added a unit test for it in the latest force push that I also verified failed prior to this change.

tests/test_async_recorder.py
```diff
@@ -43,6 +44,28 @@ async def test_capture(loop):
...
+                # Ensure that the task subsegments overlap
+                if counter < total_tasks:
+                    await event.wait()
+                else:
+                    event.set()
+                return subsegment.parent_id
+        tasks = [assert_task() for task in range(total_tasks)]
+        results = await asyncio.gather(*tasks)
+        for result in results:
+            assert result == segment.id
```
NathanielRN, line 67: Just a suggestion on naming:

```suggestion
        subsegs_parent_ids = await asyncio.gather(*tasks)
        for subseg_parent_id in subsegs_parent_ids:
            assert subseg_parent_id == segment.id
```

NathanielRN (approved):
Thanks so much for your help Tyler! This is a great bug fix and I look forward to seeing it being deployed 🙂 

I added some suggestions for documentation to help us in the future, would you mind taking a look? Thanks!

tyler-dodge:
The modifications you added all look great to me.

Issue #212, opened 2020-04-03
future dependency should be scoped to python 2 in setup.py


kapilt:
currently there's an unconditional dep on the future library but afaics its just compatibility to python3, ideally that should be detected in the handful of places its used to avoid the shim/alias on python3. afaics there's only two uses

./aws_xray_sdk/core/plugins/ec2_plugin.py
2:from future.standard_library import install_aliases

./aws_xray_sdk/ext/sqlalchemy/util/decorators.py
4:from future.standard_library import install_aliases

PR #343, opened 2022-06-23
Only import future for py2

NathanielRN:
*Issue #, if available:* #212

*Description of changes:*

If we only import `future` when the SDK is run on Python 2, then we don't need to import it for `Python 3` systems. This will prevent conflicts with people downloading conflicting versions of the `future` package.

Fixes: #212

### DIFF
--- a/aws_xray_sdk/core/plugins/ec2_plugin.py
+++ b/aws_xray_sdk/core/plugins/ec2_plugin.py
@@ -1,7 +1,11 @@
 import json
 import logging
-from future.standard_library import install_aliases
-install_aliases()
+
+from aws_xray_sdk.core.utils.compat import PY2
+
+if PY2:
+    from future.standard_library import install_aliases
+    install_aliases()
 
 from urllib.request import urlopen, Request
 
--- a/aws_xray_sdk/ext/sqlalchemy/util/decorators.py
+++ b/aws_xray_sdk/ext/sqlalchemy/util/decorators.py
@@ -3,8 +3,13 @@ import types
 
 from aws_xray_sdk.core import xray_recorder
 from aws_xray_sdk.ext.util import strip_url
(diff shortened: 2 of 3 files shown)


aws_xray_sdk/core/plugins/ec2_plugin.py
```diff
@@ -1,7 +1,10 @@
 import json
 import logging
-from future.standard_library import install_aliases
-install_aliases()
+try:
+    from future.standard_library import install_aliases
+    install_aliases()
+except ImportError:
+    pass
```
srprash, line 7: Instead of using `try-except`, can we reuse the [`PY2`](https://github.com/aws/aws-xray-sdk-python/blob/d025f5641496efdc9dfad2b270139c224053885c/aws_xray_sdk/core/utils/compat.py#L5) variable from `compat.py`?

Then this could be something like:
```python
from aws_xray_sdk.core.utils.compat import PY2
    
if PY2:
    from future.standard_library import install_aliases
    install_aliases()
```

NathanielRN: Oh interesting! I took this path because [StackOverflow says it was the status quo](https://stackoverflow.com/a/3496790/7460739) and I saw it [used in OpenTelemetry Python Contrib](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/0007c9046c74d724c342c769bca2a874357be907/instrumentation/opentelemetry-instrumentation-django/src/opentelemetry/instrumentation/django/middleware.py#L46-L52), but this `PY2` looks super useful so I'll change it to that :)

srprash: I see. That's good to know. :)
Since we already have the other mechanism, I think it would be better to just reuse for the sake of consistency.

PR #348, opened 2022-07-25
Add annotations

millarm:
This enables searching by URL values for implementations of Django in Lambda working around the restriction that http.url can't be searched on subsegments in the X-Ray Console.


carolabadeer (commented):
Hi!

We already support filtering / searching HTTP attributes by using string keywords. Please see the _filter expressions_ section in the [X-Ray developer guide documentation](https://docs.aws.amazon.com/xray/latest/devguide/xray-console-filters.html#console-filters-string). 

You can use `http.url BEGINSWITH "https://your.url.com"` to achieve the same result for your use case.

millarm:
> You can use `http.url BEGINSWITH "https://your.url.com"` to achieve the same result for your use case.

Unfortunately not....   (this took 4 weeks to resolve with AWS support)

This only works where the root segment has had the metadata.   As you'll see from the code when running in a Lambda environment the Xray SDK creates subsegments for each request (which handles the re-use of lambdas)

This means... it's currently not possible to filter with http.url on Xray deployed in AWS Lambda.

Hence this PR -> this uses the workaround recommended by AWS support to use the annotations, which are filterable on subsegments.

While I agree that filtering on http.url would be the ideal solution, in the absence of a change in the way Xray filtering is implemented, this workaround resolves the problem of "I can't filter at all" for anyone deploying in a Lambda environment, while leaving the http.url filtering unchanged for those deploying outside of Lambda.

carolabadeer:
I see, that is definitely a valid point and a change we will consider. However, since X-Ray imposes a limit of 50 annotations per trace, it is important that adding these annotations is an opt-in feature as not to use up the allocated amount for others who may not necessarily need it for this use case. 

This can be done by adding a new configuration to the Django middleware. You can find more information on how to add one [here](https://github.com/aws/aws-xray-sdk-python#add-django-middleware).

millarm:
> since X-Ray imposes a limit of 50 annotations per trace

Thanks, that's a good point, and one I wasn't aware of.

I 💯 agree that this can be worked around with a custom Django middleware (this is the solution we have deployed to production today)

However I'm raising this PR as it seems... daft to force everyone to:

1. Discover this limitation that's unique to lambda
2. Have to create their own middleware to... do the same thing

So I feel it would be better if:

a. This was supported "out of the box" by the AWS SDK
b. Configured on by default (so that people don't waste time going "oh, my lambda deployed XRay isn't searchable by URL"
  (though maybe I'm channeling the fact it took us 12 months to prioritise this then 4 weeks with AWS support to resolve the problem!)
  
So... final update to this PR where I take these comments into account incoming.

millarm:
@carolabadeer see update -> which matches my desire to "avoid people getting the wrong default behaviour" and allows smart users to configure for either ALL or NONE as appropriate.

aws_xray_sdk/ext/django/middleware.py
```diff
@@ -50,9 +59,10 @@ def __call__(self, request):
             recorder=xray_recorder,
             sampling_req=sampling_req,
         )
-
         if self.in_lambda_ctx:
-            segment = xray_recorder.begin_subsegment(name)
+            segment = xray_recorder.begin_subsegment(name, namespace="remote")
```
carolabadeer, line 63: Can you please clarify the reason for setting `namespace="remote"` here? As far as I know, setting a subsegment's namespace as remote manually makes it show on the service map and should not affect the ability to search annotations.

millarm: Helpfully directly from the Xray team here:

> change the namespace of the subsegments from "local" to "remote" and to then use the service filter. When the namespace is set to "remote", Xray will infer a node base which can be searched up with the service filter.

So this enables service filter use in this case too

carolabadeer: I referred back to our team and confirmed that we do index all annotations in a trace, which includes annotations added on subsegments. In this case, it doesn't seem necessary to set `namespace="remote"` in order to be able to search annotations, so we can merge this PR once that adjustment is made!

millarm: Thank you!   Updated with removing this.    Will plan to remove our patches once this is released, and looking forward to nobody else hitting this issue!

carolabadeer (approved):

Issue #354, opened 2022-09-19
Add raise from syntax to discover underlying problems


lehmat:
# Issue
AWS SDK does not follow `raise exc from exc` which causes underlying problems to be logged to be rooted from AWS SDK instead of the underlying issue. Reference: [Pynamodb issue](https://github.com/pynamodb/PynamoDB/issues/1073)

# Proposed change
Follow [PEP 3134](https://peps.python.org/pep-3134/) to raise underlying issues as traceback instead of stopping it to AWS SDK

PR #355, opened 2022-09-19
Implement PEP3134 to discover underlying problems with python3

lehmat:
*Issue #, if available:*
#354 

*Description of changes:*
Implement PEP3134 through [six.raise_from](https://six.readthedocs.io/index.html?highlight=raise_from#six.raise_from) to support python2 and python3.

Also move one import to top of the module to better comply with pylint (C0413:wrong-import-position)

### DIFF
--- a/aws_xray_sdk/ext/sqlalchemy_core/patch.py
+++ b/aws_xray_sdk/ext/sqlalchemy_core/patch.py
@@ -1,12 +1,13 @@
 import logging
 import sys
+import wrapt
+import six
 
 if sys.version_info >= (3, 0, 0):
     from urllib.parse import urlparse, uses_netloc
 else:
     from urlparse import urlparse, uses_netloc
 
-import wrapt
 
 from aws_xray_sdk.core import xray_recorder
 from aws_xray_sdk.core.patcher import _PATCHED_MODULES
@@ -72,12 +73,12 @@ def _process_request(wrapped, engine_instance, args, kwargs):
         subsegment = None
     try:
         res = wrapped(*args, **kwargs)
-    except Exception:
+    except Exception as exc:
         if subsegment is not None:
             exception = sys.exc_info()[1]
             stack = stacktrace.get_stacktrace(limit=xray_recorder._max_trace_back)
             subsegment.add_exception(exception, stack)
-        raise
+        six.raise_from(exc, exc)
     finally:
         if subsegment is not None:
             subsegment.set_sql(sql)
--- a/aws_xray_sdk/core/async_recorder.py
+++ b/aws_xray_sdk/core/async_recorder.py
@@ -1,4 +1,5 @@
 import time
+import six
 
 from aws_xray_sdk.core.recorder import AWSXRayRecorder
 from aws_xray_sdk.core.utils import stacktrace
@@ -81,10 +82,10 @@ class AsyncAWSXRayRecorder(AWSXRayRecorder):
         try:
             return_value = await wrapped(*args, **kwargs)
             return return_value
-        except Exception as e:
-            exception = e
+        except Exception as exc:
+            exception = exc
             stack = stacktrace.get_stacktrace(limit=self._max_trace_back)
-            raise
+            six.raise_from(exc, exc)
         finally:
             # No-op if subsegment is `None` due to `LOG_ERROR`.
             if subsegment is not None:
--- a/aws_xray_sdk/core/patcher.py
+++ b/aws_xray_sdk/core/patcher.py
@@ -6,6 +6,7 @@ import pkgutil
 import re
 import sys
 import wrapt
+import six
 
 from aws_xray_sdk import global_sdk_config
 from .utils.compat import PY2, is_classmethod, is_instance_method
@@ -107,9 +108,9 @@ def patch(modules_to_patch, raise_errors=True, ignore_module_patterns=None):
 def _patch_module(module_to_patch, raise_errors=True):
     try:
         _patch(module_to_patch)
-    except Exception:
+    except Exception as exc:
         if raise_errors:
-            raise
+            six.raise_from(exc, exc)
         log.debug('failed to patch module %s', module_to_patch)
 
 
--- a/aws_xray_sdk/core/recorder.py
+++ b/aws_xray_sdk/core/recorder.py
@@ -4,6 +4,7 @@ import logging
 import os
 import platform
 import time
+import six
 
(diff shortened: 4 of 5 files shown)


aws_xray_sdk/ext/sqlalchemy_core/patch.py
```diff
@@ -1,12 +1,13 @@
...
 import sys
+import wrapt
+import six
 
 if sys.version_info >= (3, 0, 0):
     from urllib.parse import urlparse, uses_netloc
 else:
     from urlparse import urlparse, uses_netloc
 
-import wrapt
```
lehmat, line 9: Just moved upwards to remove pylint:C0413

lehmat:
This ticket is tied to the following troubleshooting and discovering this issue: [Pynamodb 1073](https://github.com/pynamodb/PynamoDB/issues/1073)

carolabadeer (approved):
LGTM!

michael-k:
@lehmat I've never seen `raise e from e` where both the exception and the cause are the same.  When playing with that in IPython, I don't see any benefit either.  Instead it adds a lot of noise:

```python
Python 3.11.0 (main, Nov 12 2022, 07:10:51) [GCC 11.2.1 20220219]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.6.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: try:
   ...:     raise ValueError("foo")
   ...: except ValueError as exc:
   ...:     print("bar")
   ...:     raise exc from exc
   ...:
bar
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
    [... skipping hidden 1 frame]

Cell In [1], line 5
      4 print("bar")
----> 5 raise exc from exc

Cell In [1], line 2
      1 try:
----> 2     raise ValueError("foo")
      3 except ValueError as exc:

ValueError: foo

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
    [... skipping hidden 1 frame]

Cell In [1], line 5
      4 print("bar")
----> 5 raise exc from exc

Cell In [1], line 2
      1 try:
----> 2     raise ValueError("foo")
      3 except ValueError as exc:

ValueError: foo

The above exception was the direct cause of the following exception:

ValueError                                Traceback (most recent call last)
Cell In [1], line 5
      3 except ValueError as exc:
      4     print("bar")
----> 5     raise exc from exc

Cell In [1], line 2
      1 try:
----> 2     raise ValueError("foo")
      3 except ValueError as exc:
      4     print("bar")

ValueError: foo
```

PEP 3134 doesn't mention such a use case either.
Can you elaborate?

<details><summary>raise</summary>

```python
In [2]: raise ValueError("foo")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In [2], line 1
----> 1 raise ValueError("foo")

ValueError: foo
```

</details>

<details><summary>try-catch-raise without from</summary>

```python
In [3]: try:
   ...:     raise ValueError("foo")
   ...: except ValueError as exc:
   ...:     print("bar")
   ...:     raise exc
   ...:
bar
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In [3], line 5
      3 except ValueError as exc:
      4     print("bar")
----> 5     raise exc

Cell In [3], line 2
      1 try:
----> 2     raise ValueError("foo")
      3 except ValueError as exc:
      4     print("bar")

ValueError: foo
```

</details>

lehmat:
This PR was made because there was underlying issue hidden by the missing raise from syntax in aws-xray-sdk core. If this would have been implemented previously, the underlying issue would have been seen without disabling xray in the application code.

### What was seen in logs prior to this PR
* https://github.com/pynamodb/PynamoDB/issues/1073

### What this PR improves
* Before: The root cause is found at `File "aws_xray_sdk/core/recorder.py", line 436, in record_subsegment`,  with invalid quote_from_bytes arguments
* After: The root cause if found at ` File sentry_sdk/tracing_utils.py", line 541` with invalid quote_from_bytes argument

Essentially it will give that `The above exception was the direct cause of the following exception:` to be pointing at underlying issue at sentry-sdk instead of stopping the traceback at aws_xray_sdk

Reference of the underlying issue that was actually the issue:
* https://github.com/getsentry/sentry-python/issues/1626

michael-k:
I saw both issues and read through them, but that doesn't explain *why* this is an issue in the first place.  A simple re-raise should not omit parts of the stacktrace.  In pynamodb/PynamoDB#1073 you said that you “disable[d] AWS X-ray to get the proper root cause to popup.”  Did you also try it with this patch and X-Ray enabled?  I'm not convinced that this actually solves your problem.

Maybe X-Ray [accessing the stacktrace](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/utils/stacktrace.py) has something to do with it?  I played around with that, but found no hint that would support this theory.  But my toy stacktraces were small.

From [`raise`'s documentation](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement):

> Changed in version 3.11: If the traceback of the active [except](https://docs.python.org/3/reference/compound_stmts.html#except)ion is modified in an except clause, a subsequent raise statement re-raises the exception with the modified traceback. Previously, the exception was re-raised with the traceback it had when it was caught.

getsentry/sentry-python#1626 mentions Python 3.9, therefore the re-raise shouldn't have changed the traceback.

lehmat:
After reviewing your comments I agree with you that this does not solve the problem. 

> Did you also try it with this patch and X-Ray enabled?

No, manipulating dependencies would be troublesome with our way of deploying lambda

Essentially this would behave as it did (with sentry-sdk[flask]==1.9.5)

```
import os
import serverless_wsgi

import sentry_sdk
from flask import Flask
import boto3
from aws_xray_sdk.core import patch_all

xray_recorder.configure(service='Test')
patch_all()


app = Flask(__name__)


sentry_sdk.init(
    dsn=os.getenv("SENTRY_DSN"),  # need to have some sane here to actually make sentry work
    release=1,  # This was the breaking on sentry-sdk, integer release causing it
    traces_sample_rate=1,
)

@app.route("/hello")
def hello_world():
    segment = xray_recorder.begin_segment('segment_name')
    boto3.client('sts').get_caller_identity().get('Account')
    xray_recorder.end_segment()
    return "<p>Hello, World!</p>"

def handler(event, context):
    return serverless_wsgi.handle_request(app, event, context)
```

PR #356, opened 2022-09-26
Fix UnboundLocalError when aiohttp server raises a CancelledError

armisael:
*Issue #, if available:* N/A

*Description of changes:* In `aiohttp` middleware, catch `BaseException` instead of `Exception` to avoid `UnboundLocalError`.

### DIFF
--- a/aws_xray_sdk/ext/aiohttp/middleware.py
+++ b/aws_xray_sdk/ext/aiohttp/middleware.py
@@ -67,7 +67,7 @@ async def middleware(request, handler):
         # Non 2XX responses are raised as HTTPExceptions
         response = exc
         six.raise_from(exc, exc)
-    except Exception as exc:
+    except BaseException as exc:
         # Store exception information including the stacktrace to the segment
         response = None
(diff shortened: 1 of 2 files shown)


carolabadeer (commented):
Looks good so far. Can you please provide a code snippet that causes an `UnboundLocalError` to be raised just for reference?

armisael:
I'm not sure, unfortunately. We have a proxy to some external services, to serve as a cache; such proxy uses aiohttp, and connects to both REST (with httpx) and SOAP (with zeep) services; we recently upgraded from python 3.6 to 3.10 and this error happened once; it's still not clear to me what exactly caused the `CancelledError`.

carolabadeer (approved):

carolabadeer (changes requested):
Merged main into this branch and now some CI tests are failing - can you please take a look?

carolabadeer:
@srprash thank you for pointing that out! Looks good to merge now that the issue is fixed.
@armisael sorry for the confusion

carolabadeer (approved):

Issue #248, opened 2020-11-19
Support httpx/httpcore


smcoll:
i'd be interested in seeing support for [`httpx`](https://www.python-httpx.org/) alongside `aiohttp`, `httplib`, and `requests`.  The default transport layer for `httpx` is [`httpcore`](https://www.encode.io/httpcore/) which supports HTTP/2 in addition to HTTP/1.1.

These libraries have async support for `asyncio`, `trio`, and `curio`.

srprash:
Hi @smcoll Thanks for opening this feature request. I'll put this in our backlog and would work on it as soon as possible. We would highly appreciate if you want to contribute towards the `httpx` support in the SDK. :)

I would also recommend you to try and open this request on the [OpenTelemetry for Python](https://github.com/open-telemetry/opentelemetry-python) since it already has a wide array of growing [instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/master/instrumentation) support and also active contribution from the open source community. We are also actively working to provide X-Ray support for customers using OpenTelemetry Python.

smcoll:
Cool - i added issue https://github.com/open-telemetry/opentelemetry-python-contrib/issues/263

thehesiod:
this xray dinosaur has awoken from his slumber :) We just swapped to async APIs and I need it for aiohttp client tracing to figure something out so I'll be working on support, I am responsible for the ddtrace aiohttp client impl so I should be able to get 'er done shortly :)

thehesiod:
btw for others that want to enable it globally I'm doing something like this (note untested):

```python
def _wrap_clientsession_init(func, instance, args, kwargs):
    trace_config = aws_xray_trace_config()
    kwargs['trace_configs'] = [trace_config]
    return func(*args, **kwargs)

wrapt.wrap_function_wrapper(aiohttp, "ClientSession.__init__", _wrap_clientsession_init)
```

PR #357, opened 2022-09-27
Instrument httpx >= 0.20

michael-k:
*Issue #, if available:* aws/aws-xray-sdk-python#248

*Description of changes:* Instrument httpx >= 0.20

The tests are the same as for requests.  That should be a good fit as httpx “[aims to be broadly compatible with the requests API](https://www.python-httpx.org/compatibility/)”.

### DIFF
--- a/aws_xray_sdk/ext/httpx/patch.py
+++ b/aws_xray_sdk/ext/httpx/patch.py
@@ -2,7 +2,7 @@ import httpx
 
 from aws_xray_sdk.core import xray_recorder
 from aws_xray_sdk.core.models import http
-from aws_xray_sdk.ext.util import UNKNOWN_HOSTNAME, inject_trace_header
+from aws_xray_sdk.ext.util import inject_trace_header, get_hostname
 
 
 def patch():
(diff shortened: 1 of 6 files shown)


aws_xray_sdk/ext/httpx/patch.py
```diff
@@ -0,0 +1,85 @@
...
+                http.URL,
+                str(request.url.copy_with(password=None, query=None, fragment=None)),
+            )
+
+            if return_value is not None:
+                subsegment.put_http_meta(http.STATUS, return_value.status_code)
+            elif exception:
+                subsegment.add_exception(exception, stack)
+
+        inject_trace_header(request.headers, xray_recorder.current_subsegment())
```
srprash, line 47: `inject_trace_header` should be called after a new subsegment has been created for this request, otherwise a wrong `Parent` will be injected to the request. Please correct me if I'm wrong, here (and also in the `handle_async_request`), it looks trace header is injected before a new subsegment is created.

aws_xray_sdk/ext/httpx/patch.py
```diff
@@ -0,0 +1,85 @@
...
+        return xray_recorder.record_subsegment(
+            wrapped=self._wrapped_transport.handle_request,
+            instance=self._wrapped_transport,
+            args=(request,),
+            kwargs={},
+            name=request.url.host or UNKNOWN_HOSTNAME,
```
srprash, line 53: you can probably use the `get_hostname` method from util to get hostname from `request.url`. The method implicitly returns either the actual hostname or the `UNKNOWN_HOSTNAME` if it cannot.

srprash (commented):
Overall it looks good to me, except for few small callouts. Thanks for contributing this new instrumentation! :)

michael-k:
The test failures are unrelated.  [flask-sqlalchemy](https://pypi.org/project/flask-sqlalchemy/#history) 3.0 was released with many breaking [changes](https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/changes/).

srprash (approved):
Thanks!

PR #366, opened 2022-11-08
Oversampling Mitigation

carolabadeer:
*Issue #, if available:* N/A

*Description of changes:* Implemented oversampling mitigation to support creating subsegments without sampling (control sampling decisions at the subsegment level) and an SQS message helper class. 

Note: Since it is possible to modify the `sampled` flag of a subsegment without the need to create a new API, this is the design implemented in the changes below. For instance, to create a new _unsampled_ subsegment and add it to a Lambda facade segment: 
```
    segment = FacadeSegment('facade_segment', 'id', 'id', True)
    subsegment = Subsegment('sampled_subsegment', 'local', segment)
    subsegment.sampled = False
    segment.add_subsegment(subsegment)
```

Similarly, the `begin_subsegment` API has been modified to include a default `sampling` parameter which defaults to `True`, and can be set to `False` when creating a new unsampled subsegment. 
```
    segment = xray_recorder.begin_segment('segment')
    subsegment = xray_recorder.begin_subsegment('unsampled_subsegment', sampling=False)
```

### DIFF
--- a/aws_xray_sdk/core/models/entity.py
+++ b/aws_xray_sdk/core/models/entity.py
@@ -81,6 +81,7 @@ class Entity(object):
         """
         self._check_ended()
         subsegment.parent_id = self.id
+        subsegment.sampled = self.sampled if subsegment.sampled == True else False
         self.subsegments.append(subsegment)
 
     def remove_subsegment(self, subsegment):
--- a/aws_xray_sdk/core/models/entity.py
+++ b/aws_xray_sdk/core/models/entity.py
@@ -81,7 +81,10 @@ class Entity(object):
         """
         self._check_ended()
         subsegment.parent_id = self.id
-        subsegment.sampled = self.sampled if subsegment.sampled == True else False
+
+        if not self.sampled and subsegment.sampled:
+            log.warning("This sampled subsegment is being added to an unsampled parent segment/subsegment and will be orphaned.")
+
         self.subsegments.append(subsegment)
 
     def remove_subsegment(self, subsegment):
--- a/aws_xray_sdk/core/recorder.py
+++ b/aws_xray_sdk/core/recorder.py
@@ -274,6 +274,20 @@ class AWSXRayRecorder(object):
             return entity.parent_segment
         else:
             return entity
+    
+    def _begin_subsegment_helper(self, name):
+        '''
+        Helper method to begin_subsegment and begin_subsegment_without_sampling
+        '''
+        result = True
+        if not global_sdk_config.sdk_enabled():
+            result =  DummySubsegment(DummySegment(global_sdk_config.DISABLED_ENTITY_NAME))
+
+        segment = self.current_segment()
+        if not segment:
+            log.warning("No segment found, cannot begin subsegment %s." % name)
+            result =  None
+        return result, segment
 
     def begin_subsegment(self, name, namespace='local'):
         """
@@ -288,52 +302,35 @@ class AWSXRayRecorder(object):
         # Generating the parent dummy segment is necessary.
         # We don't need to store anything in context. Assumption here
         # is that we only work with recorder-level APIs.
-        if not global_sdk_config.sdk_enabled():
-            return DummySubsegment(DummySegment(global_sdk_config.DISABLED_ENTITY_NAME))
-
-        segment = self.current_segment()
-        if not segment:
-            log.warning("No segment found, cannot begin subsegment %s." % name)
-            return None
+        result, segment = self._begin_subsegment_helper(name)
+        if type(result) == DummySubsegment or result == None:
+            return result
 
         current_entity = self.get_trace_entity()
 
         if not current_entity.sampled:
             subsegment = DummySubsegment(segment, name)
-            subsegment.sampled = False
         else:
             subsegment = Subsegment(name, namespace, segment)
 
         self.context.put_subsegment(subsegment)
-
         return subsegment
 
     def begin_subsegment_without_sampling(self, name):
         """
-        Begin a new subsegment.
+        Begin a new unsampled subsegment.
         If there is open subsegment, the newly created subsegment will be the
         child of latest opened subsegment.
         If not, it will be the child of the current open segment.
 
         :param str name: the name of the subsegment.
-        :param str namespace: currently can only be 'local', 'remote', 'aws'.
         """
-        # Generating the parent dummy segment is necessary.
-        # We don't need to store anything in context. Assumption here
-        # is that we only work with recorder-level APIs.
-        if not global_sdk_config.sdk_enabled():
-            return DummySubsegment(DummySegment(global_sdk_config.DISABLED_ENTITY_NAME))
-
-        segment = self.current_segment()
-        if not segment:
-            log.warning("No segment found, cannot begin subsegment %s." % name)
-            return None
+        result, segment = self._begin_subsegment_helper(name)
+        if type(result) == DummySubsegment or result == None:
+            return result
 
         subsegment = DummySubsegment(segment, name)
-        subsegment.sampled = False
- 
         self.context.put_subsegment(subsegment)
-
         return subsegment
 
     def current_subsegment(self):
--- a/aws_xray_sdk/core/recorder.py
+++ b/aws_xray_sdk/core/recorder.py
@@ -274,20 +274,32 @@ class AWSXRayRecorder(object):
             return entity.parent_segment
         else:
             return entity
-    
-    def _begin_subsegment_helper(self, name):
+
+    def _begin_subsegment_helper(self, name, namespace='local', beginWithoutSampling=False):
         '''
         Helper method to begin_subsegment and begin_subsegment_without_sampling
         '''
-        result = True
+        # Generating the parent dummy segment is necessary.
+        # We don't need to store anything in context. Assumption here
+        # is that we only work with recorder-level APIs.
         if not global_sdk_config.sdk_enabled():
-            result =  DummySubsegment(DummySegment(global_sdk_config.DISABLED_ENTITY_NAME))
+            return DummySubsegment(DummySegment(global_sdk_config.DISABLED_ENTITY_NAME))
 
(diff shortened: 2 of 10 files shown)


carolabadeer:
This change has been tested in a Lambda environment using the following Lambda function:
```
from aws_xray_sdk.core.models.subsegment import Subsegment
from aws_xray_sdk.core.utils.sqs_message_helper import SqsMessageHelper
import logging 

def lambda_handler(event, context):
    
    for message in event['Records']:
        if SqsMessageHelper.isSampled(message):
            print('sampled - doing batch work')
            subsegment = xray_recorder.begin_subsegment('sampled_subsegment', sampling=True)

        else:
            print('unsampled - doing batch work')
            subsegment = xray_recorder.begin_subsegment('unsampled_subsegment', sampling=False)
    
    xray_recorder.end_subsegment()    
```

X-Ray service map when `Sampled=1` in `AWSTraceHeader`:
image (7)

X-Ray service map when `Sampled=0` in `AWSTraceHeader`:
image (8)

aws_xray_sdk/core/models/entity.py
```diff
@@ -81,6 +81,7 @@ def add_subsegment(self, subsegment):
         """
         self._check_ended()
         subsegment.parent_id = self.id
+        subsegment.sampled = self.sampled if subsegment.sampled == True else False
```
srprash, line 84: Why are we doing parent based sampling if the `subsegment` itself is sampled? Do we need this line at all?

aws_xray_sdk/core/recorder.py
```diff
@@ -284,6 +284,8 @@ def begin_subsegment(self, name, namespace='local'):
 
         :param str name: the name of the subsegment.
         :param str namespace: currently can only be 'local', 'remote', 'aws'.
+        :param bool sampling: sampling decision for the subsegment being created,
+            defaults to True
```
srprash, line 288: This `sampling` parameter is not the same in meaning with the `sampled` flag on an entity.
`sampling=True` here means that the subsegment will inherit the sampling decision of its parent entity, and could either be sampled or not. 
`sampling=False` would mean that the subsegment will not inherit the sampling decision and will be unsampled.

I hope I was able to explain the difference. Please reword the description. :)

aws_xray_sdk/core/recorder.py
```diff
@@ -335,7 +340,7 @@ def end_subsegment(self, end_time=None):
 
         # if segment is already close, we check if we can send entire segment
         # otherwise we check if we need to stream some subsegments
-        if self.current_segment().ready_to_send():
+        if self.current_segment().ready_to_send():
```
srprash, line 343: nit: unnecessary white space.

aws_xray_sdk/core/utils/sqs_message_helper.py
```diff
@@ -0,0 +1,10 @@
+class SqsMessageHelper:
+    
+    @staticmethod 
+    def isSampled(sqs_message):
+        attributes = sqs_message['attributes']
+
+        if 'AWSTraceHeader' not in attributes:
```
srprash, line 7: I suggest making `AWSTraceHeader` into a constant and reuse it in both the places in this method.

aws_xray_sdk/ext/util.py
```diff
@@ -35,7 +35,7 @@ def inject_trace_header(headers, entity):
     else:
         header = entity.get_origin_trace_header()
     data = header.data if header else None
-
+
```
srprash, line 38: nit: unnecessary white spaces

aws_xray_sdk/core/recorder.py
```diff
@@ -296,8 +298,11 @@ def begin_subsegment(self, name, namespace='local'):
             log.warning("No segment found, cannot begin subsegment %s." % name)
             return None
 
-        if not segment.sampled:
+        current_entity = self.get_trace_entity()
+
+        if not current_entity.sampled or not sampling:
```
atshaw43, line 303: Would looking at just current_entity.sampled be enough here?

atshaw43:
Can we use the same paradigm of a separate without sampled function instead of passing in a parameter? We want the SDKs to be consistant.

atshaw43:
Can we also get a unit test for the emitter?

srprash:
@carolabadeer I know we decided on this approach in our conversation. But now when I think of it again, I agree with @atshaw43 that we should keep the SDKs consistent. There could be confusion among customers why the a `begin_subsegment_without_sampling` API exists in other SDKs but not in this.

atshaw43:
> @carolabadeer I know we decided on this approach in our conversation. But now when I think of it again, I agree with @atshaw43 that we should keep the SDKs consistent. There could be confusion among customers why the a `begin_subsegment_without_sampling` API exists in other SDKs but not in this.

aws_xray_sdk/core/models/entity.py
carolabadeer: This is specifically for the case when a subsegment is added to an unsampled subsegment. We have to ensure that once an unsampled subsegment is created, all of its child subsegments are also unsampled. 

So for example, it is possible to do this:
```
    segment = Segment('parent_segment', 'id', 'id', True)
    subsegment = Subsegment('subsegment1', 'local', segment)
    subsegment.sampled = False
    segment.add_subsegment(subsegment)
    subsegment2 = Subsegment('subsegment2', 'local', segment)
    subsegment.add_subsegment(subsegment2)
```
In this case, it would be illegal for `subsegment2` to be sampled, so it has to take the decision of it's direct parent subsegment - `subsegment1`- which is unsampled.

aws_xray_sdk/core/recorder.py
carolabadeer: I don't think so. If the current entity (segment or subsegment) is sampled and a new unsampled subsegment is being added, it is necessary to check the sampling flag being set on the newly created subsegment.

carolabadeer:
Updated lambda handler function with the new `begin_subsegment_without_sampling` API:
```
def lambda_handler(event, context):

    for message in event['Records']:
        if SqsMessageHelper.isSampled(message):
            print('sampled - doing batch work')
            subsegment = xray_recorder.begin_subsegment('sampled_subsegment')

        else:
            print('unsampled - doing batch work')
            subsegment = xray_recorder.begin_subsegment_without_sampling('unsampled_subsegment')
    
    xray_recorder.end_subsegment()    
```

aws_xray_sdk/core/models/entity.py
atshaw43: Does it make more sense to move this assignment to the "begin_subsegment" function in recorder.py and just base it off of the parent there? Then we can just base it off of the entity here "current_entity = self.get_trace_entity()". That entity would be the direct parent no?

atshaw43: Having to check two levels like that is a code smell to me.

carolabadeer: There are two different ways to create subsegments, which is why this had to be done on two different levels - @srprash please correct me if I'm missing something

aws_xray_sdk/core/recorder.py
```diff
@@ -296,15 +296,46 @@ def begin_subsegment(self, name, namespace='local'):
             log.warning("No segment found, cannot begin subsegment %s." % name)
             return None
 
-        if not segment.sampled:
+        current_entity = self.get_trace_entity()
+
+        if not current_entity.sampled:
             subsegment = DummySubsegment(segment, name)
+            subsegment.sampled = False
```
srprash, line 303: This code is redundant. A `DummySubsegment` is already unsampled.
https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/dummy_entities.py#L96

aws_xray_sdk/core/recorder.py
```diff
@@ -296,15 +296,46 @@ def begin_subsegment(self, name, namespace='local'):
...
         else:
             subsegment = Subsegment(name, namespace, segment)
 
         self.context.put_subsegment(subsegment)
 
         return subsegment
 
+    def begin_subsegment_without_sampling(self, name):
```
srprash, line 311: There's lot of duplicate code here with `begin_subsegment`. Can you move the common code to another function and reuse?

aws_xray_sdk/core/recorder.py
```diff
@@ -296,15 +296,46 @@ def begin_subsegment(self, name, namespace='local'):
...
+        if not global_sdk_config.sdk_enabled():
+            return DummySubsegment(DummySegment(global_sdk_config.DISABLED_ENTITY_NAME))
+
+        segment = self.current_segment()
+        if not segment:
+            log.warning("No segment found, cannot begin subsegment %s." % name)
+            return None
+
+        subsegment = DummySubsegment(segment, name)
+        subsegment.sampled = False
```
srprash, line 333: Need not set `sampled = False` on a `DummySubsegment`.

atshaw43 (approved):

aws_xray_sdk/core/recorder.py
```diff
@@ -274,6 +274,20 @@ def current_segment(self):
             return entity.parent_segment
         else:
             return entity
+    
+    def _begin_subsegment_helper(self, name):
+        '''
+        Helper method to begin_subsegment and begin_subsegment_without_sampling
+        '''
+        result = True
```
srprash, line 282: I feel this is still problematic. I'm not sure I get the point of using `result` and returning it.
Also, `result` seem to be of variable types. Here it is a boolean but later it can be a Subsegment. This can easily confusion and bugs.

I believe there is a better way to leverage this helper method than it's callers having to check its return types and values.
`_begin_subsegment_helper` can accept a parameter as flag for sampling or not. The callers can simply call this method with the flag and get back a Subsegment depending on the flag's value. How does that sound?
The method definition could look like:
```python
def _begin_subsegment_helper(name, namespace='local', beginWithoutSampling=False)
```

PR #372, opened 2022-12-01
Modified context missing strategy default to log error

carolabadeer:
*Issue #, if available:* N/A

*Description of changes:* To maintain consistency across SDKs and to prevent malformed trace headers from breaking services, this PR sets the default context missing strategy to `'LOG_ERROR'` instead of `'RUNTIME_ERROR'`.

### DIFF
--- a/tests/test_recorder.py
+++ b/tests/test_recorder.py
@@ -13,6 +13,7 @@ from aws_xray_sdk import global_sdk_config
 from aws_xray_sdk.core.models.segment import Segment
 from aws_xray_sdk.core.models.subsegment import Subsegment
 from aws_xray_sdk.core.models.dummy_entities import DummySegment, DummySubsegment
+from aws_xray_sdk.core.exceptions.exceptions import SegmentNotFoundException
 
 xray_recorder = get_new_stubbed_recorder()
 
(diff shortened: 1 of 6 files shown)


tests/test_recorder.py
```diff
@@ -99,19 +100,26 @@ def test_put_annotation_metadata():
...
 
     xray_recorder.put_annotation('key', 'value')
     xray_recorder.put_metadata('key', 'value')
     xray_recorder.end_segment()
 
+def test_raise_runtime_error_with_missing_context():
+    xray_recorder = get_new_stubbed_recorder()
+    xray_recorder.configure(sampling=False, context_missing='RUNTIME_ERROR')
+
+    with pytest.raises(SegmentNotFoundException):
```
atshaw43, line 116: I'm not too familiar with Python. If there is not an exception throw, as expected, then will this test fail?

carolabadeer: It will! This is what happens when you modify the `context_missing` parameter to be `LOG_ERROR` in this test:
`Failed: DID NOT RAISE <class 'aws_xray_sdk.core.exceptions.exceptions.SegmentNotFoundException'>`

srprash: Adding to Carol's comment above, this is a common way in pytest to assert that a certain exception is thrown in a unit test. Documentation for reference: https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

srprash (approved):
LGTM!

PR #376, opened 2023-01-17
Run tests against Python 3.10 and 3.11

michael-k:
*Issue #, if available:* n/a

*Description of changes:* Run tests against Python 3.10 and 3.11.  Also add trove classifiers.

----

### DIFF
--- a/.github/workflows/UnitTesting.yaml
+++ b/.github/workflows/UnitTesting.yaml
@@ -21,10 +21,12 @@ jobs:
       py37: 3.7
       py38: 3.8
       py39: 3.9
+      py310: '3.10'
+      py311: '3.11'
     strategy:
       fail-fast: false
       matrix:
-        python-version: [py27, py34, py35, py36, py37, py38, py39]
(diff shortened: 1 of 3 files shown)


michael-k:
The one failing test is unrelated.

I did however not expect for the tests to pass because the packages used to provide PostgreSQL and MySQL aren't compatible with Python 3.1x.  But at least the ones used for MySQL have to be replaced by others anyway when moving away from Ubuntu 18.04 (which is deprecated by GitHub Actions).  But that's out of scope for this PR.

srprash:
Looking at the latest CI run, seems like we'll have to fix the sqlalchemy tests before we can add python 3.1x versions to it.
We will be working on fixing the CI shortly.

srprash:
I will go ahead and approve this PR since the test failures are unrelated to this change. I will create another PR in the meantime to pin the sqlalchemy to 1.x.
Thanks!

srprash (approved):

PR #381, opened 2023-03-14
version pinning sqlalchemy to 1.x to fix CI

srprash:
*Issue #, if available:*
Follow up from https://github.com/aws/aws-xray-sdk-python/pull/376#issuecomment-1408035455

*Description of changes:*
Pinning the sqlalchemy dependency to < 2.0.0 since there are breaking changes in the new major version, rendering it incompatible with the X-Ray SDK.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -88,10 +88,11 @@ deps =
 
     ext-flask_sqlalchemy: flask >= 0.10
     ext-flask_sqlalchemy: Flask-SQLAlchemy <= 2.5.1
+    ext-flask_sqlalchemy: sqlalchemy >=1.0.0,<2.0.0
 
-    ext-sqlalchemy: sqlalchemy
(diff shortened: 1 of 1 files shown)


srprash:
Well, now the py35 and py39 tests are failing for the `requests` instrumentation.

willarmiros:
Can we remove some older Python versions like 3.5 and 3.6 as they're EOL? Or would that be considered a breaking change?

srprash:
> Can we remove some older Python versions like 3.5 and 3.6 as they're EOL? Or would that be considered a breaking change?

I think it will be a breaking change. If there are customers still using python 3.5 or 3.6 and have minor version dependency on the SDK, they will break on the next minor version release.

carolabadeer (approved):

srprash:
> Well, now the py35 and py39 tests are failing for the requests instrumentation.

Turns out these tests have become flaky.

PR #384, opened 2023-03-30
Start MySQL from GH actions, upgrade ubuntu, and remove python versions for tests

jj22ee:
*Issue #, if available:*
Ubuntu 18.04 deprecation:
https://github.com/actions/runner-images/issues/6002
Also, the latest Ubuntu image uses MySQL 8, which breaks the `testing.mysqld` dependency as it uses `mysql_install_db` which is removed from MySQL 8.

*Description of changes:*
- Bump unbuntu image to 22.04, remove python 2.7, 3.4, 3.5, 3.6 from unit test
- Start MySQL from GH actions instead of relying on `testing.mysqld` dependency

### DIFF
--- a/.github/workflows/UnitTesting.yaml
+++ b/.github/workflows/UnitTesting.yaml
@@ -9,28 +9,29 @@ on:
 
 jobs:
   test:
-    # "setup-python" action doesn't provide python 3.4 binaries for ubuntu-latest.
-    # Sticking to Ubuntu 18.04 as recommended here:
-    # https://github.com/actions/setup-python/issues/185#issuecomment-768232756
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     env:
-      py27: 2.7
-      py34: 3.4
-      py35: 3.5
-      py36: 3.6
       py37: 3.7
       py38: 3.8
       py39: 3.9
       py310: '3.10'
       py311: '3.11'
+      DB_DATABASE: test_db
+      DB_USER: root
+      DB_PASSWORD: root
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


srprash:
Should we also update the `tox.ini` to remove the unsupported python environments?

wangzlei (approved):
LGTM

jj22ee:
> Should we also update the `tox.ini` to remove the unsupported python environments?

Good catch, I will make this update

srprash (approved):

tox.ini
```diff
@@ -112,8 +98,7 @@ deps =
     ext-pg8000: testing.postgresql
 
     ext-pymysql: testing.mysqld
```
wangzlei, line 100: We don't need testing.mysqld anymore?

jj22ee: Thanks, updated!

.github/workflows/UnitTesting.yaml
```diff
@@ -9,29 +9,30 @@ on:
...
       matrix:
-        python-version: [py27, py34, py35, py36, py37, py38, py39, py310, py311]
+        python-version: [py37, py38, py39, py310, py311]
         testenv: [core, ext]
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
 
+      - name: Start MySQL
+        run: |
```
wangzlei, line 32: add 
```
if: ${{ matrix.testenv == 'ext' }}
```
between name and run

willarmiros: Protip @wangzlei , you can use suggestions for this, e.g.

```suggestion
        if: ${{ matrix.testenv == 'ext' }}
        run: |
```

PR #387, opened 2023-04-04
Cleanup after drop of support for Python < 3.7

### LINKED ISSUE #255: Deprecate support for python 2.7 and 3.4 in the next major version

michael-k:
*Issue #, if available:* Closes #255

*Description of changes:*

- Remove compatibility code for Python < 3.7
- Introduce `python_requires` and cleanup `setup.py`
- Use new style classes and `super()` calls
- Remove unused config for Travis CI


setup.py
```diff
@@ -28,23 +28,17 @@
...
-        'Programming Language :: Python :: 3.5',
-        'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',
         'Programming Language :: Python :: 3.11',
     ],
 
+    python_requires=">=3.7",
```
srprash, line 39: Nice! I believe this makes the removal of older runtimes safe for users using py <37 as they will not get the new version, right?

srprash (approved):
Thanks for going through the code and cleaning up! 🙂

setup.py
michael-k: Yes, as long as they use a recent enough release of pip (v9.0 released 2016-11-02) and the environment that generates the package metadata uses setuptools v42 (released 2019-11-23) or later.

PR #391, opened 2023-04-27
Disable IMDSv1 from Elastic Beanstalk

wangzlei:
Integration workflow verify the functionality in AWS Elastic Beanstalk. For AWS security reason we need to disable IMDSv1 from EB environment.

*Issue #, if available:*

*Description of changes:*
Screenshot 2023-04-26 at 5 10 47 PM

### DIFF
--- a/terraform/eb.tf
+++ b/terraform/eb.tf
@@ -71,4 +71,10 @@ resource "aws_elastic_beanstalk_environment" "eb_env" {
     name = "XRayEnabled"
     value = "true"
   }
+  
+  setting {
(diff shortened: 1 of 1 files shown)


carolabadeer:
Including discussion in this PR for visibility: Since some EB environments were still running and not destroyed correctly (likely due to an issue while running the integration test workflow), can we also update the workflow to ensure that the EB environments being created are always being destroyed correctly regardless of the outcome of the workflow?

carolabadeer (approved):

PR #399, opened 2023-08-07
Fix passing multiple values in testenv.passenv in tox.ini

musicinmybrain:
Current versions of tox advise: `pass_env values cannot contain whitespace, use comma to have multiple values in a single line`

*Issue #, if available:*
Fixes #398.

*Description of changes:*

Separates `testenv.passenv` entries with commas rather than spaces in `tox.ini`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #400, opened 2023-08-08
Enable testing on Python 3.12

musicinmybrain:
*Issue #, if available:*

*Description of changes:*

Added an explicitly dependency on `setuptools`, since a test imports `pkg_resources`.

Added Python 3.12 versions of all the `tox` environments that supported Python 3.11.

Added Python 3.12 to GitHub unit testing workflow.

I did *not* add Python 3.12 to the trove classifiers, but I’m happy to add this if it’s wanted.

*Additional information:*

Testing this PR requires https://github.com/aws/aws-xray-sdk-python/pull/399.

It may also be necessary to use a very recent version of `pip`:

```
python3.12 -m venv _e
. _e/bin/activate
pip install --upgrade pip tox
tox
```

Most tests pass. Some HTTP tests seem to fail or flake in all Python versions due to transient issues at https://httpbin.org/. At a glance, regressions specific to Python 3.12 appear to be due to certain dependencies not being ready.

----


musicinmybrain:
Compare to: https://github.com/aws/aws-xray-sdk-python/pull/376/

musicinmybrain:
The motivation for this PR is:

- I am [adding `aws-xray-sdk-python`](https://bugzilla.redhat.com/show_bug.cgi?id=2189323) to Fedora Linux, which already has Python 3.12 in the development version (Rawhide) and will release Fedora 39 with Python 3.12 this fall.
- Since test dependencies are generated from (Python-version-specific) `tox` environments, I needed to patch in Python 3.12 versions of the existing environments to generate the proper dependencies for the RPM build process.
- Offering necessary changes upstream is [good practice](https://docs.fedoraproject.org/en-US/package-maintainers/Staying_Close_to_Upstream_Projects/#why_push_changes_upstream).

.github/workflows/UnitTesting.yaml
```diff
@@ -16,13 +16,14 @@ jobs:
       py39: 3.9
       py310: '3.10'
       py311: '3.11'
+      py311: '3.12'
```
carolabadeer, line 19: Should this be `py312`? 
```suggestion
      py312: '3.12'
```

tox.ini
```diff
@@ -56,7 +56,10 @@ deps =
     wrapt
 
     ; Python 3.5+ only deps
-    py{37,38,39,310,311}: pytest-asyncio
+    py{37,38,39,310,311,312}: pytest-asyncio
+
+    ; For pkg_resources
+    py{37,38,39,310,311,312}: setuptools
```
carolabadeer, line 62: I am still not very clear on why this dependency is needed. You mentioned it was added since a test imports `pkg_resources`, which test are you referring to?

michael-k: Probably only “necessary” on Python 3.12 though:

> [gh-95299](https://github.com/python/cpython/issues/95299): Do not pre-install setuptools in virtual environments created with [venv](https://docs.python.org/3.12/library/venv.html#module-venv). This means that distutils, setuptools, pkg_resources, and easy_install will no longer available by default; to access these run pip install setuptools in the [activated](https://docs.python.org/3.12/library/venv.html#venv-explanation) virtual environment.

https://docs.python.org/3.12/whatsnew/3.12.html

.github/workflows/UnitTesting.yaml
```diff
@@ -16,13 +16,14 @@ jobs:
       py39: 3.9
       py310: '3.10'
       py311: '3.11'
+      py312: '3.12'
```
michael-k, line 19: FYI: Should have worked at the time with `3.12.0-beta - 3.12`.  But it's available now, see https://github.com/actions/python-versions/blob/main/versions-manifest.json

Maybe rebase to trigger a rerun?

musicinmybrain:
Hmm, I’m not sure why this is failing on Python 3.12 with `ModuleNotFoundError: No module named 'setuptools'` from `setup.py`. In local testing with `tox`, the added dependency on `setuptools` in `tox.ini` works as expected.

tox.ini
musicinmybrain: Perhaps, but it’s not harmful on others. The dependency is still real; it’s just allowed to be implicit. I guess I can change to `py312` only if implicit is better than explicit.

michael-k:
`run: pip install "tox<=3.27.1" -U tox-factor` should install `setuptools` as well.  The added dependency is only available inside the venv that tox creates.  But it's required outside as well to build the sdist.

https://github.com/aws/aws-xray-sdk-python/blob/8563afcc9a0b2591a288237122331343f260a026/.github/workflows/UnitTesting.yaml#L42-L43

musicinmybrain:
> `run: pip install "tox<=3.27.1" -U tox-factor` should install `setuptools` as well. The added dependency is only available inside the venv that tox creates. But it's required outside as well to build the sdist.

Thanks. Hopefully https://github.com/aws/aws-xray-sdk-python/pull/400/commits/2a682c933b34bb2b55813990457813a7db2b7df2 is correct and sufficient.

srprash (approved):

PR #412, opened 2023-10-19
Pin flask < 3.x for flask sqlalchemy tests

srprash:
*Issue:*
Flask SqlAlchemy under the hood tries to import '_app_ctx_stack' from Flask but this has been removed in the new flask v3.0.0: https://github.com/pallets/flask/pull/5223

*Description of changes:*
Pinning the flask dependency major version to prevent breaking the unit tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -74,7 +74,7 @@ deps =
 
     ext-flask: flask >= 0.10
 
-    ext-flask_sqlalchemy: flask >= 0.10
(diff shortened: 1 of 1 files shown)


atshaw43:
Does this library only affect unit tests?
If not, would it be better to upgrade SqlAlchemy instead of pinning Flask?

atshaw43:
When these tests start failing?
Seems odd that this kind of issue just pops up.

atshaw43 (approved):

Issue #416, opened 2023-11-20
sqlalchemy_core patch errors for unencoded special characters in db url


jachien:
## Problem

There is an [open sqlalchemy bug](https://github.com/sqlalchemy/sqlalchemy/issues/10662) where the `sqlalchemy.engine.url.URL` class does not correctly url encode username and passwords when converting to a string.

This behavior can cause the xray patching of sqlalchemy `Session` and `Connection` `execute` functions to error when trying to [urlparse the db url returned by sqlalchemy](https://github.com/aws/aws-xray-sdk-python/blob/9e6fa9f5aed3e36ec1590172f296a2c2c8666608/aws_xray_sdk/ext/sqlalchemy_core/patch.py#L17).

In my case, a closing square bracket `]` in my password leads urlparse to interpret my db url as an invalid IPv6 address, but this could probably manifest as other errors depending on the unencoded characters.

```
Error parsing sql metadata.
Traceback (most recent call last):
  File "path/to/aws_xray_sdk/ext/sqlalchemy_core/patch.py", line 22, in _sql_meta
    url = urlparse(str(engine_instance.engine.url))
  File "path/to/urllib/parse.py", line 384, in urlparse
    splitresult = urlsplit(url, scheme, allow_fragments)
  File "path/to/urllib/parse.py", line 486, in urlsplit
    raise ValueError("Invalid IPv6 URL")
ValueError: Invalid IPv6 URL
```
Note: stack trace is for aws-xray-sdk 2.8.0, the linked code above is for current master head.

## What I expected
No error logged from xray and full trace functionality.

## My setup
aws-xray-sdk 2.8.0
Aurora Postgres 14 (but can affect any database)
pg8000 1.20.0 (but can affect any dialect)
Python 3.8

## Steps to reproduce
```
CREATE DATABASE test_xray;
CREATE USER test_xray_user WITH ENCRYPTED PASSWORD 'test]password';
GRANT ALL PRIVILEGES ON DATABASE test_xray TO test_xray_user;
```

```
from sqlalchemy import create_engine
import urllib

from aws_xray_sdk.core import patch_all, xray_recorder
patch_all()

password = urllib.parse.quote_plus("test]password")
db_url = f"postgresql+pg8000://test_xray_user:{password}@myhost:5432/test_xray"

print(db_url)
# postgresql+pg8000://test_xray_user:test%5Dpassword@myhost:5432/test_xray

engine = create_engine(db_url)

[...]

PR #418, opened 2024-01-29
Fix issue 416

vastin:
*Issue #, if available:*
#416 

*Description of changes:*
- Accepted proposed solution in issue #416.
- Fixed broken unit test and integration test.


aws_xray_sdk/ext/sqlalchemy_core/patch.py
```diff
@@ -14,18 +14,21 @@
 def _sql_meta(engine_instance, args):
```
jj22ee, line 14: Can we add new test case for passwords like `test]password`?

vastin: The unit test is added.

PR #429, opened 2024-05-07
[Lambda] Create dummy segment when trace header is incomplete

majanjua-amzn:
When using Lambda, if there is no trace header or an invalid one we would like to send a NoOp segment instead of a Facade segment so that customers can call X-Ray SDK APIs and utilize X-Ray SDK without any output. This will not send data to X-Ray nor inject X-Ray trace headers downstream.

*Description of changes:*
- When initializing the context, create a dummy segment if the trace header is incomplete
- Appropriately identify when the created segment is a dummy and handle it accordingly (` or isinstance(current_entity, DummySegment)`)
- Update comments and tests accordingly

### DIFF
--- a/aws_xray_sdk/core/lambda_launcher.py
+++ b/aws_xray_sdk/core/lambda_launcher.py
@@ -3,6 +3,7 @@ import logging
 import threading
 
 from aws_xray_sdk import global_sdk_config
+from .models.dummy_entities import DummySegment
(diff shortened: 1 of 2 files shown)


wangzlei:
Do we have a unit test cover the trace header "Root=x;Lineage=y" case in Lambda enviroment?

wangzlei (approved):

PR #455, opened 2025-04-16
Modernize event loop handling in aiohttp tests

musicinmybrain:
*Issue #, if available:*

**N/A**

*Description of changes:*

This fixes compatibility with `pytest-aiohttp` 1.1.0.

First, migrate `aiohttp` tests from `loop` fixture to `event_loop`. This follows advice from an upstream deprecation warning, and furthermore, as of `pytest-aiohttp` 1.1.0 these tests break since the `loop` fixture is no longer available at all.

Then, in async tests, replace `event_loop` fixtures with calls to `asyncio.get_running_loop()`, as advised by, e.g.:

```
tests/ext/aiohttp/test_middleware.py:123
  tests/ext/aiohttp/test_middleware.py:123: PytestDeprecationWarning: test_ok is asynchronous and expl
icitly requests the "event_loop" fixture. Asynchronous fixtures and test functions should use "asyncio
.get_running_loop()" instead.
    async def test_ok(aiohttp_client, event_loop, recorder):
```

### DIFF
--- a/tests/ext/aiohttp/test_client.py
+++ b/tests/ext/aiohttp/test_client.py
@@ -17,22 +17,22 @@ BASE_URL = 'httpbin.org'
 
 
 @pytest.fixture(scope='function')
-def recorder(loop):
+def recorder(event_loop):
     """
     Initiate a recorder and clear it up once has been used.
     """
-    xray_recorder.configure(service='test', sampling=False, context=AsyncContext(loop=loop))
+    xray_recorder.configure(service='test', sampling=False, context=AsyncContext(loop=event_loop))
     xray_recorder.clear_trace_entities()
     yield recorder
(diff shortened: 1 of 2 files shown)


musicinmybrain:
Thank you for the feedback. I tried rebasing this on `master` and found there were several other cases of the `event_loop` fixture in async tests that needed to be handled. Plus, it wasn’t immediately clear to me how to deal with the `recorder` fixture you pointed out. Since updating this PR doesn’t look completely trivial, and I am no longer maintaining a [`python-aws-xray-sdk` package](https://src.fedoraproject.org/rpms/python-aws-xray-sdk) in Fedora, I’m not motivated to spend more time on this. Please take it as a useful starting point, if in fact it is one.

